PR #14, opened 2010-09-05
Added git-blame support

sandbender:
Most (not all) cli options are present here, with a few missing as well as a few hard-coded to be used (ie: --incremental and -p, since they're needed/used here for parsing, and we don't want/need pretty human readable output from the binary, etc).

No unit tests yet - I'll commit some if you end up pulling these changes (recent activity is low, so not sure if you'll see this?), but for the moment I just needed blame support in an app I'm writing, so I figured I'd add it in here instead of just hacking it into my local copy... it works so far for what I'm doing, but if you find any bugs please let me know and I'll definitely correct them asap.

Enjoy :)


Nakilon:
So there is no git blame?

jcouball:
@Nakilon no, blame support was never added.

Nakilon:
> @Nakilon no, blame support was never added.

ok, I used gem rugged

Issue #33, opened 2011-10-14
Crash if log contains foreign characters(?)


fwal:
Hi!

I'm running version 1.2.5 of the gem and 1.9.2 (p290) of ruby.
The following code crashes if the log contains a foreign character in the git log:

``` ruby
require 'git'

# Init git and get log
g = Git.init
log = g.log

changelog = "My changes: \n"

# Now create changelog (and crash)
log.each {|l| 
   changelog += "- " + l.message + "\n")
}


```

Stack trace:

```
[...]/gems/git-1.2.5/lib/git/lib.rb:672:in `split': invalid byte sequence in US-ASCII (ArgumentError)
   from [...]/gems/git-1.2.5/lib/git/lib.rb:672:in `command_lines'
   from [...]/gems/git-1.2.5/lib/git/lib.rb:92:in `full_log_commits'
   from [...]/gems/git-1.2.5/lib/git/log.rb:108:in `run_log'
   from [...]/gems/git-1.2.5/lib/git/log.rb:101:in `check_log'
   from [...]/gems/git-1.2.5/lib/git/log.rb:84:in `each'
   from gittest:10:in `<main>'
```

Any ideas?

fwal:
Seems abandoned 🗑

ut:
Looking at the queue of the waiting PR, that's a pity.

rvodden:
Myself at @perlun are in the process of resurrecting the project. I don't want this issue to get lost so I'm re-openong. We've not discussed priorities yet, however internationalization is likely to be high.

ut:
It seems that this issue has been solved in the meantime. I succesfully checked it with the example provided by @fwal

fwal:
That's awesome @ut, then I'll close this for real 🎉

PR #54, opened 2012-11-28
fix utf8 issue + remove deprecated has_rdoc

ut:
i fixed the utf8 issue (which occurs if non-latin1 chars are used in a git commit message) with a solution described in issue [#33](https://github.com/schacon/ruby-git/issues/33#issuecomment-2838152)  and by the way i removed a deprecated syntax for rdoc

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -1,34 +1,34 @@
 require 'tempfile'
 
 module Git
-  
-  class GitExecuteError < StandardError 
+
+  class GitExecuteError < StandardError
   end
-  
+
   class Lib
-      
+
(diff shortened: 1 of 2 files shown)


perlun:
Thank @ut. Would you have a chance to rebase this on top of the current `master`? So we can see if it is CI clean etc.

ut:
Thanks @perlun. I checked my PR: The fix provided has been solved in the meantime by commit  fd86ddd93 in reference to issue #190. I've verfied the fix by using the example in #33, which worked for me. You might consider to close #33 and #190 also.

perlun:
Alright, great! Thanks for the confirmation. Closing this now.

Issue #82, opened 2013-06-30
Feature request: Support option `--follow` for logging


icy:
The native git command has the --follow option to see logs of files that are moved / renamed.

Could you please take a look and support an option for the .log operation, for example

```
$repo.object($object).log(10, true)
#10: number of log entriess
# true: continue listing the history of a file beyond renames
```

Thank you

icy:
The feature is implemented in https://github.com/ruby-git/ruby-git/pull/93. This issue can be closed. Thanks a lot @skade

PR #93, opened 2013-08-31
Implement Git::Log#follow

skade:
This implements Git::Log#follow, mentioned in #82.

In contrast to the issue description, it implements the follow option as a seperate method instead of a parameter.

The implementation doesn't check for sanity (`--follow` is only allowed for single file logs).

Example:

```
@repo.log.path(file_path).follow
@repo.log.path(file_path).follow(true).follow(false) #off again
```

Tests are in a seperate commit. There is now indication how manipulating the test suite git repo should be done, I just went ahead with what I think how it works. Feel free to remove the commit.


evantravers:
Did this get removed? I can't see a reference to it in the documentation anymore… and testing in pry seems to indicate it's gone.

jcouball:
@evantravers I don't see that this ever made it into master.

PR #140, opened 2014-03-19
added feature to create orphan branches

srse:
This pull request also contains two fixes: one for test execution under Ruby 2.1.1 and one for the usage of '-b' used in lib.checkout. Just merge the orphan support commits, if you do not want to merge these fixes. I added/extended the test cases for all the changes I have made.

By the way: when will be the next release date, so that I can switch back from my custom gem to the rubyforge-release as soon as the feature to create orphan branches is merged?

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -522,7 +522,7 @@ module Git
     def checkout(branch, opts = {})
       arr_opts = []
       arr_opts << '-f' if opts[:force]
-      arr_opts << '-b' << opts[:new_branch] if opts[:new_branch]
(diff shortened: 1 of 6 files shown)


lib/git/lib.rb
```diff
@@ -521,11 +521,21 @@ def branch_delete(branch)
     
     def checkout(branch, opts = {})
       arr_opts = []
+      arr_opts << '--orphan' if opts[:orphan]
       arr_opts << '-f' if opts[:force]
-      arr_opts << '-b' << opts[:new_branch] if opts[:new_branch]
```
robertodecurnex, line 525: Good catch

PR #151, opened 2014-06-17
Update author.rb

ACOKing:
This pull request will make rgit more server-like and will give it networking functionality

### DIFF
--- a/lib/git/author.rb
+++ b/lib/git/author.rb
@@ -1,12 +1,13 @@
 module Git
   class Author
-    attr_accessor :name, :email, :date
+    attr_accessor :name, :email, :date, :ipaddress
     
     def initialize(author_string)
       if m = /(.*?) <(.*?)> (\d+) (.*)/.match(author_string)
         @name = m[1]
         @email = m[2]
         @date = Time.at(m[3].to_i)
+        @ipaddress = m[4].to_s
       end
     end
     


lib/git/author.rb
```diff
@@ -1,14 +1,15 @@
 module Git
   class Author
-    attr_accessor :name, :email, :date
+    attr_accessor :name, :email, :date, :ipaddress
```
ACOKing, line 3: This IP Address property allows  an IP address to be specified in rgit

lib/git/author.rb
```diff
@@ -1,14 +1,15 @@
...
     
     def initialize(author_string)
       if m = /(.*?) <(.*?)> (\d+) (.*)/.match(author_string)
         @name = m[1]
         @email = m[2]
         @date = Time.at(m[3].to_i)
+        @ipaddress = m[4].to_s
```
ACOKing, line 10: This line converts the IP address to a string for human readability

ghost, line 10: `m[4]` is regexp matched string so there is no need it 'to_s' method.
Furthermore, author_string does not contain any ip addresses.
This string originates from author/committer lines of `git log --format=raw` command and looks like this:

`Scott Chacon <schacon@gmail.com> 1252081154 -0700`

So, what's the point of your changes?

PR #152, opened 2014-06-17
Custom git binary support

spikegrobstein:
Provide support for specifying a custom `git` executable. Since we need to ensure that our omnibussed `git` is executed without mucking around with my PATH, there is now a `@@git_binary` variable in `Git::Lib` which can be set via `Git::Lib.git_binary = '/path/to/git'`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,12 @@ Require the 'git' gem.
     require 'git'
 ```
 
+This gem will use the first `git` that it finds in your `PATH` (i.e. when you type `git` on the commandline), if you require a specific executable, you can override it by doing something like the following:
+
+```ruby
+    Git::Lib.git_binary = '/path/to/git'
+```
+
 Here are the operations that need read permission only.
 
 ```ruby


lib/git/lib.rb
```diff
@@ -742,7 +753,7 @@ def command(cmd, opts = [], chdir = true, redirect = '', &block)
 
       opts = [opts].flatten.map {|s| escape(s) }.join(' ')
 
-      git_cmd = "git #{cmd} #{opts} #{redirect} 2>&1"
+      git_cmd = "#{@@git_binary} #{cmd} #{opts} #{redirect} 2>&1"
```
robertodecurnex, line 756: I would avoid the usage of the class variable here and use the method #get_binary method instead.

lib/git/lib.rb
```diff
@@ -6,7 +6,10 @@ class GitExecuteError < StandardError
   end
   
   class Lib
-      
+
+    DEFAULT_GIT_BINARY = 'git'
+    @@git_binary = DEFAULT_GIT_BINARY
```
robertodecurnex, line 11: Should be deleted if we move this into the getter.

lib/git/lib.rb
```diff
@@ -25,6 +28,14 @@ def initialize(base = nil, logger = nil)
       @logger = logger
     end
 
+    def self.git_binary=(new_git_binary)
+      @@git_binary = new_git_binary
+    end
+
+    def self.git_binary
+      @@git_binary
```
robertodecurnex, line 36: We can lazy-load this var using the CONT if not defined. `@@git_binary ||= DEFAULT_GIT_BINARY`

README.md
```diff
@@ -57,6 +57,12 @@ Require the 'git' gem.
     require 'git'
 ```
 
+This gem will use the first `git` that it finds in your `PATH` (i.e. when you type `git` on the commandline), if you require a specific executable, you can override it by doing something like the following:
+
+```ruby
+    Git::Lib.git_binary = '/path/to/git'
```
robertodecurnex, line 63: There are a couple of changes like this on the pipeline. It may be a good time to add a single configuration entry.

Thinking in something like:

```
Git.config do |config|
  config.binary_path = 'path/to/binary'
  config.ssh_keys = 'path/to/key'
end
```

spikegrobstein, line 63: I really like that style as it's a better implementation than what I put together.

PR #160, opened 2014-07-16
pass env variables to popen, rather than setting them globally

thisissmithj:
The Git library sets global state when running instance methods, which is suboptimal and unexpected. The issue is important, as it affects not only other instantiated Git objects, but also affects simple `git blah` commands. That makes debugging very difficult.

This is a better solution to the global environment variables issue than the other pull request that I saw.  If you set then unset global variables, you're going to have issues in multi-threaded programs (I had issues when using celluloid worker threads). However, if you put the env data into a hash and pass it to each popen call, then there is no requirement for global state at all.

Also note that all of the methods that I changed are private.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -1,3 +1,4 @@
+require 'open3'
 require 'tempfile'
 
 module Git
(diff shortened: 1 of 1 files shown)


robertodecurnex:
cc95cfa77489b383a98c388a30a20d1d8f373549 is a fixing most of the problems. We can re-discuss the usage of pipes but not under the current implementation. Feel free to open an issue, as feature, if you still need more chenges in the way env variables are used (a7b30d997382aa08cf4d79fc2139596210ee5419 is also introducing mutex to prevent problems with threads).

thisissmithj:
That looks good, thanks. Do you have an ETA for a new version release?

robertodecurnex:
Trying to close 1.2.9 during this week.

PR #167, opened 2014-08-10
Enable mirror option for git clone and push.

larsxschneider:
c.f. https://help.github.com/articles/duplicating-a-repository

Tested and used here:
https://github.com/larsxschneider/hubsync


arpad9:
this would be really nice.  what needs to happen here?  do you need more tests? @robertodecurnex

guimaluf:
Why is this was abandoned ? Since 2014, this is a must have option to git clone lib!

PR #173, opened 2014-08-26
Quick fix

T145:
Hope it helps! :palm_tree:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,11 +1,10 @@
+require 'git/version'
 require 'rubygems'
-
-require "#{File.expand_path(File.dirname(__FILE__))}/lib/git/version"
 
 task :default => :test
 
 desc 'Run Unit Tests'
-task :test do |t|
+task :test do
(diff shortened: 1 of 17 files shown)


git.gemspec
```diff
@@ -12,36 +15,18 @@ Gem::Specification.new do |s|
...
-    'lib/git/path.rb',
-    'lib/git/remote.rb',
-    'lib/git/repository.rb',
-    'lib/git/stash.rb',
-    'lib/git/stashes.rb',
-    'lib/git/status.rb',
-    'lib/git/version.rb',
-    'lib/git/working_directory.rb'
-  ]
+  s.files = `git ls-files -z`.split("\x0")
```
robertodecurnex, line 29: I'm a public enemy of this particular line. It makes the spec useless after the .gem is pushes. In the other side there's obviously not a git repo to get that information from.

I know it's a very usual practice, I have engaged into discussions hunderd of times, bundler gem made this a default, so sad :(

PR #177, opened 2014-09-14
Keep original timezone in author/committer date

ppsshh:
Keep original commit timezone instead leaving it undefined (local)


jamietanna:
While updating https://github.com/ivantsepp/jekyll-git_metadata/issues/6 I found that the ability to handle timezones is required - would you accept a PR to add this functionality?

PR #189, opened 2014-11-09
fix that Git#checkout will work with :new_branch

detlefw:
this small fix, will makes Git#checkout work if :new_branch is given as an argument, like

git = Git.open 'path/to/repo'
git.checkout 'branchName', :new_branch => true

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -522,7 +522,7 @@ module Git
     def checkout(branch, opts = {})
       arr_opts = []
       arr_opts << '-f' if opts[:force]
-      arr_opts << '-b' << opts[:new_branch] if opts[:new_branch]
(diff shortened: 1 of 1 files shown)


lib/git/lib.rb
```diff
@@ -522,7 +522,7 @@ def branch_delete(branch)
     def checkout(branch, opts = {})
       arr_opts = []
       arr_opts << '-f' if opts[:force]
-      arr_opts << '-b' << opts[:new_branch] if opts[:new_branch]
+      arr_opts << '-b' << if opts[:new_branch]
```
robertodecurnex, line 525: This is syntactically invalid, the last `<<` must be removed.

Anyway, based on the PR comments some refactor should be done in order to make it backwards compatible.

PR #209, opened 2015-01-03
Adding branch option support for #clone.

NotDaveLane:
This makes it possible for developers to call Git.clone and include a `branch` in the `opts` hash.


robertodecurnex:
Added some tests over that. Ty @NotDaveLane !

PR #210, opened 2015-01-03
Change mentions of ssh_key to ssh_wrapper

mikedillion:
This [commit](https://github.com/schacon/ruby-git/commit/92c088b88ae25812c96afda38d5a7e7e030e82b0) was a little misleading. The `GIT_SSH` environment variable let's Git know where to find an SSH wrapper, not the path for a key to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Git env config
     config.binary_path = '/git/bin/path'
 
     # If you need to use a custom SSH Key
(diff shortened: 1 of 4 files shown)


robertodecurnex:
`config.git_ssh` (just the name of the ENV variable) sounds like a cleaner choise. Ty @mikedillion ! 55dae73f49d434690ff34ddd86738ae9beb70680

PR #212, opened 2015-01-13
fix git ssh env usage

cf-pub-tools:
v1.2.8 allowed setting of GIT_SSH to point to SSH wrapper.
v1.2.9 overrides this with value of Git::Base.config.git_ssh.

We are not 100% on the v1.2.8 behavior - whether the config setting or the ENV var should take precedence. 
For our purposes, the ENV var would be preferable.

This is our story in our backlog: https://www.pivotaltracker.com/story/show/86059268.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -866,7 +866,7 @@ module Git
       ENV['GIT_DIR'] = @git_dir
       ENV['GIT_WORK_TREE'] = @git_work_dir
       ENV['GIT_INDEX_FILE'] = @git_index_file
-      ENV['GIT_SSH'] = Git::Base.config.git_ssh
+      ENV.fetch('GIT_SSH', Git::Base.config.git_ssh)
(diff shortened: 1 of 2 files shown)


robertodecurnex:
My bad. It should probably default to the current ENV value if there is no explicit call to `Git::Config.git_ssh=`
--
Robertt

-----Original Message-----

robertodecurnex:
@cf-pub-tools Can you please confirm if the repo's master version works for you?

Ty!

robertodecurnex:
Adding tests from PR - 14738bba33463e62cb43d52c2b5d9b7f6a1b3474

camelpunch:
I've just verified that v1.2.9.1 works in the context where v1.2.9 didn't (I was also working on the pull request with @ericTsiliacos and fkotsian). Thanks for fixing and releasing this - we can now un-pin our dependency!

PR #217, opened 2015-02-06
Allow a block to receive push output

bsedat:
This allows for an optional block to receive a pipe that it can print to stdout (useful when deploying to Heroku for example).

### DIFF
--- a/tests/units/test_remotes.rb
+++ b/tests/units/test_remotes.rb
@@ -190,5 +190,25 @@ class TestRemotes < Test::Unit::TestCase
     end
   end
 
+  def test_push_with_block
+    in_temp_dir do |path|
+      loc = Git.clone(@wbare, 'local')
+      rem = Git.clone(@wbare, 'remote', :config => 'receive.denyCurrentBranch=ignore')
 
+      loc.add_remote('testrem', rem)
+
+      loc.chdir do
+        new_file('test-file', 'blahblahblah')
+        loc.add
+        loc.commit('master commit')
+      end
+
+      block_called = false
+      loc.push('testrem') do |output|
+        assert_match(/master -> master/, output.read, 'expected block to receive output of push command.')
+        block_called = true
+      end
+      assert(block_called, 'block was not called.')
+    end
+  end
 end
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -322,11 +322,11 @@ module Git
     #
     #  @git.config('remote.remote-name.push', 'refs/heads/master:refs/heads/master')
     #
-    def push(remote = 'origin', branch = 'master', opts = {})
+    def push(remote = 'origin', branch = 'master', opts = {}, &block)
       # Small hack to keep backwards compatibility with the 'push(remote, branch, tags)' method signature.
       opts = {:tags => opts} if [true, false].include?(opts)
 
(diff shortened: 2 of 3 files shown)


robkinyon-roivant:
Bump

tarcinil:
@robkinyon-roivant 

What is your use-case for this?  I would like to see some testing added to this.  While it didn't break any previous tests, it doesn't show any testing around the concept that it is adding.

robkinyon-roivant:
@tarcinil I have a few, all around providing a wrapper around Git for developers.

1. Provide a place to debug failures, particularly with failures of assumption. For example, if I have setup a branch that's mirrored to a remote and attempt to push to it, but someone else deleted the remote branch in the meantime. There should be a place to catch the output so I can parse what Git provides so my wrapper can tell the user the next step to take.
1. Provide a way to inject explanatory text into the output. This is especially useful when building a teaching tool that's distributed for computer-assisted instruction.

These reasons aren't limited to push, but I figured it would be easier to start the discussion in the specific rather than the general.

bsedat:
I added a quick test - let me know if you'd like stricter assertions.

robkinyon-roivant:
@tarcinil - Stupid question: Given that

- `command()` takes a block and passes it to `run_command()` which passes it to `IO.popen()`
- All the commands eventually get down to `command()`

Why not, as a matter of course, just have all command take an optional block which tramps down to `IO.popen()`?

tests/units/test_remotes.rb
```diff
@@ -190,5 +190,25 @@ def test_push
...
+        new_file('test-file', 'blahblahblah')
+        loc.add
+        loc.commit('master commit')
+      end
+
+      block_called = false
+      loc.push('testrem') do |output|
+        assert_match(/master -> master/, output.read, 'expected block to receive output of push command.')
+        block_called = true
+      end
```
perlun, line 210: I am not convinced, yet. Why not just let the `push` method return the output? Does it already return something different? (I admit, I haven't looked into the details.)

robkinyon-roivant: As I said to @tarcinil, given that `command()` takes an optional block, why shouldn't all invokers of `command()` pass along an optional block?

perlun: @robkinyon-roivant Sure, it would be possible. Would you care to write a PR for it, and also add some practical real-world examples of why and when it would be useful?

If it's _only_ for people writing tests, I am a bit more hesitant to make the change than if it's for general usability of the gem. I am not saying "no", I am just saying "maybe". All added code makes things more complex => each change have to be weighed towards "is it worth the  [potentially] decreased readability and [potentially] increased complexity", that's all.

bsedat: `Git::Lib#command` already returns the output of the command, so `push` would as well. The block here eventually gets passed down to `IO.popen` and provides access to the stdout stream of the git command as it runs. 
We use this library for pushing changes to a remote PaaS git server that runs some lifecycle hooks on an incoming push and it was useful to be able to view output as it happens rather than just getting a timeout (and potentially losing the output data).

perlun: Fair enough, I'm willing to accept this @bsedat @robkinyon-roivant.

Please submit a PR that fixes it for all invokers of `command`, or was this already the case?

Issue #239, opened 2015-07-19
Git checkout doesn't track remote


BillyBarbaro:
When checking out a new branch, you cannot set it to track the remote version of the branch.  I've forked the repo and added the --track option to the checkout method.  I plan to submit a pull request, but will wait a few days to make sure there's no additional functionality I may need to add.

echernyavskiy:
Is there any update on this effort please? Would've been really useful.

Also, to accommodate repositories with multiple remotes, any chance you could add an option to specify the remote to checkout the branch from? I might have enough knowledge of Ruby to add the option myself but not to cover that with tests unfortunately.

thierryxing:
Just fork and add --track option in [https://github.com/thierryxing/ruby-git](https://github.com/thierryxing/ruby-git)

PR #240, opened 2015-07-20
Adding track option to checkout

BillyBarbaro:
This is the feature addition I discussed in Issue #239.  Added the --track of -t flag to the checkout command.

I also rearranged the order of the arguments in that method so that both the force and branch flags could be used in the same commands.

I've added unit tests to for all the changes I made.


aleksey-hariton:
+1

When this changes will be merged?

BillyBarbaro:
Hi @aleksey-hariton.  I went ahead and pulled in the most recent version of ruby-git to my branch.  I'm not all that familiar with Ruby, so as you'll see I have a couple of checks failing on the Travis CI Build.  I'm pretty sure the failure in 1.9.2 is present in almost all the Pos, but I'm not about the jruby one.

That being said, I'm not sure about the status of the gem anymore after reading Issue #307.

PR #258, opened 2015-11-12
Add Git push --set-upstream

olleolleolle:
Additional option was added to the `push` wrapper.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -722,6 +722,7 @@ module Git
       
       arr_opts = []
       arr_opts << '--force'  if opts[:force] || opts[:f]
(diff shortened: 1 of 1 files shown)


LoicAG:
@perlun @tarcinil Could you please take a look at this PR, and possibly re-open / merge it?
It would be very useful to do the equivalent of:
```git push -u origin my_branch```

tarcinil:
@LoicAG - Is it possible that you can add a test that would test to ensure that the upstream is set?

Lightly based on https://github.com/ruby-git/ruby-git/blob/master/tests/units/test_remotes.rb#L18

jcouball:
Closing due to no test.  Happy to accept if a test is added.

Bluelich:
Are there any update of this pr?

PR #261, opened 2015-11-24
unsetter

icambron:
Provides support for unsetting configs, both locally and globally.

### DIFF
--- a/lib/git.rb
+++ b/lib/git.rb
@@ -42,35 +42,15 @@ end
 # Author::    Scott Chacon (mailto:schacon@gmail.com)
 # License::   MIT License
 module Git
-  
-  #g.config('user.name', 'Scott Chacon') # sets value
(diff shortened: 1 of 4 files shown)


lib/git.rb
```diff
@@ -42,25 +42,9 @@
 # Author::    Scott Chacon (mailto:schacon@gmail.com)
 # License::   MIT License
 module Git
-  
-  #g.config('user.name', 'Scott Chacon') # sets value
```
icambron, line 46: So this is a bit out of scope, but this method is confusing as hell. I don't see anywhere that mixes this in, it's not tested, and it's duplicative of code in `Base`.

lib/git.rb
```diff
@@ -69,10 +53,6 @@ def self.config
     return Base.config
   end
 
-  def global_config(name = nil, value = nil)
```
icambron, line 72: Same here.

PR #273, opened 2016-01-26
fixed tests on MacOSX

everm1nd:
on OSX `tar` utility prefix each line of verbose output with 'x'
that fails archive tests
fixed using `assert_match` instead of `assert_equal` for `tar` output
also `tar` outputs in STDERR on Mac so used pipeline redirect "2>&1" for capture


everm1nd:
Tests are failing with some versions of Ruby, but it looks like TravisCI misconfiguration for me. Can you assist?

PR #286, opened 2016-03-23
Incorrect information on usage in README.md

patrickviet:
You need g.object to fetch information about a commit.
g.commit will do a commit.


patrickviet:
no actualy it was gcommit. my bad.

PR #288, opened 2016-03-27
Support a couple of tag related git operations

horiaradu:


horiaradu:
Ho @schacon any news on this?

Issue #295, opened 2016-06-07
Not support Chinese character


zesming:
I usually get some mojibake if the output information (e.g. `g.log.first.author.name`) contains some Chinese character, just like '������' .

If I annotate the encoding code (`lib.rb` line 826 to 866), output would be correct (ruby 2.2.3).

``` ruby
  def command_lines(cmd, opts = [], chdir = true, redirect = '')
      cmd_op = command(cmd, opts, chdir)
    #   op = cmd_op.encode("UTF-8", "binary", {
      #     :invalid => :replace,
        # :undef => :replace
      # })
      cmd_op.split("\n")
  end
```

I have found [an issue and a pull request](https://github.com/schacon/ruby-git/issues/188) to add this implementation. Does it a good choice to convert UTF-8 manually?

PR #301, opened 2016-07-21
Fix encoding of command output

cschramm:
With the current solution and on Ruby 2.2 all non-ASCII are replaced even if they are valid UTF-8. I guess that solution works on Ruby 1.9 as the default encoding changed in 2.0, but I did not try that.

It looks like I'm not the only one experiencing the issue, see e.g. #295.


cschramm:
For reference: A similar fix got merged in #327.

PR #304, opened 2016-08-05
Fix miss encode trouble

u-ichi:
sample code

``` ruby
require 'rubygems'
require 'git'
git = Git.clone('https://github.com/u-ichi/retry-handler', 'retry-handler')
puts git.log.first.message
```

original

```
Exception���������StandardError������������������
```

this patch

```
ExceptionよりもStandardErrorのほうが適切
```


rickmzp:
👍

PR #315, opened 2016-12-02
commit with custom author date

mgarciaisaia:
Honor `git commit --date=...` flag.


krezka:
I'd love these function! Now I can set author, but date. Any chance to get these?
It'd be also great to have a way to pass custom options to commit (for case like that)

mgarciaisaia:
I can't see the failed builds logs, nor rebuild it.

Can someone retry it? I may try to fix the issues if they're there.

mgarciaisaia:
@schacon bump (?)

PR #316, opened 2016-12-04
Fix space prefix in tag message

ddfreyne:
Passing a message when creating a tag erroneously prefixes the message with a space. This PR fixes that issue.


ddfreyne:
@perlun Hi, as you seem to be maintaining ruby-git, could you take a look at this PR?

PR #318, opened 2016-12-07
Assume strings are utf8 by default and fallback on encoding if necessary

kamaradclimber:
Before this patch, we always assumed the string was ascii-8bit which
does not work correctly with real utf8 strings (french accented chars or
chineese chars).

This patch tries to use the string as utf8 and fallback if necessary to
the the ascii-8bit assumption if necessary.

This should help with #295 and might replace the need for #301.

Change-Id: Idac63fa10e5aefafa1eb99a6be4138cac5f90ea0


kamaradclimber:
Sorry for the removal of ruby 1.9.2 without prior discussion. At the moment tests cannot pass (even in master) for ruby 1.9.2 (see my comment on my second commit).

kamaradclimber:
@rvodden would you consider this PR if I fixed the conflicts?

kamaradclimber:
@perlun I am willing to fix the conflicts if you are interested in the PR. let me know

perlun:
@kamaradclimber I think #327 supersedes this. Could you try with the latest master and see if this assumption is right? If not, we can discuss it further - I'm open to changes that fixes bugs in the implementation.

kamaradclimber:
thanks for the answer. I'll close the PR and test latest master

PR #323, opened 2017-01-07
Added support for git clone '--no-single-branch' option.

jamesridgway:


jamesridgway:
I have updated this, and this is now passing on Travis. Could I get some approvals on this so this can get merged?

PR #325, opened 2017-01-19
Add support for '--delete' in push

mkody:
I needed to delete a remote branch. And the way to do it is by using the `--delete` argument in the `git push` command [(as seen in the Git docs)](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches#_delete_branches).

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -755,6 +755,7 @@ module Git
 
       arr_opts = []
       arr_opts << '--mirror'  if opts[:mirror]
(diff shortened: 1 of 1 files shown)


robkinyon-roivant:
Bump. Could this be re-prioritized?

tarcinil:
@mkody Please add DCO to your comments, and we will look at merging this.

mkody:
There we go, rebased for DCO.

PR #327, opened 2017-04-10
Fix UTF-8 message errors

drakmail:
Without this fix I got questions instead chars:

"������������������ ������������"

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -859,10 +859,14 @@ module Git
 
     def command_lines(cmd, opts = [], chdir = true, redirect = '')
       cmd_op = command(cmd, opts, chdir)
-      op = cmd_op.encode("UTF-8", "binary", {
-	  	:invalid => :replace,
-		:undef => :replace
-	  })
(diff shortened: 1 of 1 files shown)


barakwei:
👍

voxik:
This seems to be similar to #318

voxik:
And #304

voxik:
And #301

perlun:
@voxik I think perhaps auto-closing _PRs_ is a bit too aggressive. Issues is one thing, but open PRs should perhaps be exempt from auto-closing; we don't have _that_ many PRs at the moment so I think manual action to go through them is better than auto-closing.

Thanks for letting me know.

perlun:
@voxik I think perhaps auto-closing _PRs_ is a bit too aggressive. Issues is one thing, but open PRs should perhaps be exempt from auto-closing; we don't have _that_ many PRs at the moment so I think manual action to go through them is better than auto-closing.

Thanks for letting me know.

drakmail:
Thank you for merge!

PR #348, opened 2018-02-24
adding boolean functions for git status

### LINKED ISSUE #346: RFC: create boolean functions for git status
At present, we have changed, added, deleted, untracked methods but no boolean methods for checking the members of the Enumerable.

Example:
```ruby
def changed
  @files.select { |k, f| f.type == 'M' }
end
```

Proposed:
```
def changed?(file)
  changed.member?(file)
end
```

This will allow interaction based on whether a file is changed as expected or not.

```
if changed?(file.rb)
  # start things
end

### PR DESCRIPTION
**Pull request by tarcinil**
fixes #346

### DIFF
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -36,7 +36,7 @@ module Git
       @files.select { |_k, f| f.untracked }
     end
 
-    def untracked?
+    def untracked?(file)
       untracked.member?(file)
     end
     
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -1,48 +1,102 @@
 module Git
-  
+  #
+  # A class for git status
+  #
   class Status
     include Enumerable
-    
+
     def initialize(base)
       @base = base
       construct_status
     end
-    
+
+    #
+    # Returns an Enumerable containing files that have changed from the
+    # git base directory
+    #
+    # @return [Enumerable]
     def changed
       @files.select { |_k, f| f.type == 'M' }
     end
 
+    #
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tarcinil** [comment]
The test that failed is outside the scope of the change that I made.  I will look into forcing a Travis run to see if it can be replicated twice.

File: lib/git/status.rb [hunk 1]
```diff
@@ -9,19 +9,35 @@ def initialize(base)
...
+      deleted.member?(file)
     end
     
     def untracked
-      @files.select { |k, f| f.untracked }
+      @files.select { |_k, f| f.untracked }
+    end
+
+    def untracked?(file)
+      untracked.member?(file)
```
perlun: For the new methods added, would you mind adding YARD-style comments that describes briefly:

- What they do
- What the expected types of the input and return values are

The new methods added in this PR are:

- changed?
- added?
- deleted?
- untracked?

Something like this will do:

```ruby
#
# Determines whether the given file has been added to the repository
#
# @param file [String] The name of the file.
# @return [Boolean]
```

This way, it is clearer for readers of the code what the intent of these (public API methods) is.

tarcinil:
@perlun 

I took your comments to heart and while I was adding the documentation, I fixed the class so that it would pass rubocop.

```bash
vburton@vburton-VirtualBox:~/Code/ruby-git$ rubocop lib/git/status.rb
Inspecting 1 file
.

1 file inspected, no offenses detected
```

The construct_status was broken up to lower method complexity while creating methods to be responsible for untracked, modified, and added

PR #350, opened 2018-02-28
Fix ls-files for git 2.16

### LINKED ISSUE #345: read git data faild "empty string is not a valid pathspec"
读取Git数据失败git '--git-dir=/Users/dongzhao/Works/QCloud/Src/iOS/Git/QCloudiOSCodes/.git' '--work-tree=/Users/dongzhao/Works/QCloud/Src/iOS/Git/QCloudiOSCodes' ls-files '--stage' ''  2>&1:fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths

rafamanzo:
Since GIT version 2.16.0 empty pathspecs as alias to "everything" has
been made illegal.

A more detailed study this change's impact has been made at:

https://github.com/ruby-git/ruby-git/issues/345#issuecomment-369260500

Closes #345

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -396,6 +396,7 @@ module Git
     end
 
     def ls_files(location=nil)
+      location ||= '.'
       hsh = {}
       command_lines('ls-files', ['--stage', location]).each do |line|
         (info, file) = line.split("\t")


lib/git/lib.rb
```diff
@@ -396,6 +396,7 @@ def diff_index(treeish)
     end
 
     def ls_files(location=nil)
+      location ||= '.'
```
tarcinil, line 399: Why not do this in the method arguments?

```ruby
def ls_files(location='.')
...
end
```

rafamanzo: Sorry for the delay, I got no notification of your review. Thanks in advance :)

I thought about doing this, but then on [ls_remote](https://github.com/ruby-git/ruby-git/pull/350/files/965ffeea3aba0adb69627157d1ea81e7775ac004#diff-6269137dee54cb0723d82224f3c5223cL410) it was handled this way, so I kept the standard of code that was already there.

Having 2 standards seemed a bad idea to me. Also modifying ls_remote which is not subject of this PR didn't feel right.

But if you think it is not a problem modifying just here, I can change it. What do you think is best?

tarcinil: You are correct about maintaining the standards.  I had not seen `ls_remote` for my various use cases.  

👍

rafamanzo:
Thanks @tarcinil I've rebased from master so Github can FF merge. Could you have another look?

PR #342, opened 2018-01-29
Enable mirror option for git clone and push

guimaluf:
c.f. https://help.github.com/articles/duplicating-a-repository

Updating original PR #167 from
Author:    Lars Schneider <larsxschneider@gmail.com>
Date:      Sat Aug 9 23:24:41 2014 -0700

Issue #319, opened 2016-12-08
jruby tests are very flaky


kamaradclimber:
While trying to make test pass for #318, I've noticed that jruby tests are really flaky.

See https://travis-ci.org/schacon/ruby-git/builds/182204463, https://travis-ci.org/schacon/ruby-git/builds/182015600 and https://travis-ci.org/schacon/ruby-git/builds/182011837 (virtually same code).

This makes contribution hard to tests.

perlun:
I _think_ this is the same as I just saw: https://travis-ci.org/ruby-git/ruby-git/jobs/350814139

We should look into sorting this out, or disabling these tests on JRuby (if all else fails.)

jcouball:
I believe this was fixed in #403 and released in [v1.6.0.pre1](https://github.com/ruby-git/ruby-git/releases/tag/v1.6.0.pre1)

PR #353, opened 2018-03-08
Creates CONTRIBUTOR and MAINTAINERS files

### LINKED ISSUE #349: Create CONTRIBUTING.md and MAINTAINERS.md
I think it would be a good idea to add the following files

CONTRIBUTING.md
==
This will be used to detail what future contributors can expect from the project.  

These are the normal items like commits must have DCO, all existing tests should pass, and new functionality should have tests, and a peer review by @ruby-git/maintainers

MAINTAINERS.md
==
Can be used by the community to know who needs to review the pull request that are created

tarcinil:
Created to create a standard review expectation for potential contributors and a clear representation of maintainers. 

The MAINTAINERS file will help avoid issues like those created in https://github.com/ruby-git/ruby-git/pull/342#issuecomment-368945651

closes #349


tarcinil:
The failure of jruby-9.1.15.0 in Travis CI seems to be a random occurrence.  My change doesn't even change how the code works but it failed the test.  

I would propose that we allow failures of jruby-9.1.15.0 in Travis.  What do you think @perlun?

perlun:
> The failure of jruby-9.1.15.0 in Travis CI seems to be a random occurrence. My change doesn't even change how the code works but it failed the test.

You're right, it seems to be an intermittent error. I restarted the job now.

> I would propose that we allow failures of jruby-9.1.15.0 in Travis. What do you think @perlun?

I am hesitant to do that, since it would easily allow us to miss real JRuby breakage (i.e. things that for whatever reason doesn't work there.) I'm a long-time JRuby user myself, and I know how annoying it is with gems that only work on MRI, so for the time being, I think it would be much better to look into the root cause for that intermittent error. There is an issue about it: https://github.com/ruby-git/ruby-git/issues/319

CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+  * [Pull Request Requirements](#pull-request-requirements)
+  * [Code Review Process](#code-review-process)
+  * [Developer Certification of Origin (DCO)](#developer-certification-of-origin-dco)
+
+
+## How Can I Contribute?
+
+### Submitting Issues
+
+We utilize **Github Issues** for issue tracking and contributions. You can contribute in two ways:
```
perlun: Minor nit: Github -> GitHub, both here and in the other places in the document.

tarcinil:
I will rebase the branch when I get home.

CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+    person who certified (a), (b) or (c) and I have not modified
+    it.
+
+(d) I understand and agree that this project and the contribution
+    are public and that a record of the contribution (including
+    all personal information I submit with it, including my
+    sign-off) is maintained indefinitely and may be redistributed
+    consistent with this project or the open source license(s)
+    involved.
+```
```
perlun, line 94: @rvodden I'm not so sure we need the DCO. It feels a bit far-fetched to me. What do you think?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+### Your checklist for this pull request
+🚨Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.
+
+- [ ] Ensure all commits include
```
perlun: include what?

tarcinil: It should have been the DCO sign-off, but could be rendered moot by your other comments.

CONTRIBUTING.md
tarcinil: I will do tonight.

PULL_REQUEST_TEMPLATE.md
perlun: @tarcinil Fair enough. Ping me or @rvodden once you have it rebased and we'll give it another look.

CONTRIBUTING.md
rvodden: I'm with you, in that it is a bit far fetched, but on the other hand all the OS projects i work on them have them... its only a check box, and I don't think its much to ask; On the other hand if we miss it out and have no legal protection at all we risk (admittedly and extremely low risk) losing all our work. I'm minded to leave it in...

PULL_REQUEST_TEMPLATE.md
tarcinil: @perlun My branch has been rebased and comments addressed.

The thread safety issue has also struck on the Ruby 2.1.10 inside of Travis.

CONTRIBUTING.md
perlun: I agree about the legal implications, they are valid. _If_ it was only a check box, I would be fine with it, but as we have it written now:

> In the commit message of the contribution, the developer simply adds a Signed-off-by statement and thereby agrees to the DCO, which you can find below or at <http://developercertificate.org/>. 

...it means more work _every time someone makes a single commit_ and submits it to the project. (OTOH, you can automate it by using `git commit -s` or `--sign-off`, but still...)

I think, if we rephrase this to be just a mere checkbox in the PR, I am fine with it.

tarcinil: @perlun I would be hesitate to change the standard for DCO without understand the legal implications that change could introduce.  The DCO depends on the git Sign-Off-By because of the integrity that is built into Git, removing that could nullify the protections that are given to the maintainers of the project.

The IDE that I use has the ability to sign-off a commit built-in, but I do not believe that the addition of a `-s` in a command line string is asking too much either.  GitHub has tools that can check for DCO on a commit.

perlun: > The IDE that I use has the ability to sign-off a commit built-in, but I do not believe that the addition of a -s in a command line string is asking too much either. GitHub has tools that can check for DCO on a commit.

Alright, I'm willing to give it a try. We can drop this requirement later if it turns out to be too burdensome.

perlun:
@tarcinil I merged this now. Please add the tool in GitHub that checks for DCO so we don't have to check that manually.

tarcinil:
@perlun I would love to but I believe I would need to be added to the organization so that I can configure integrations. 

Please see the integration located here:
https://github.com/probot/dco

Installation can be done through:
https://github.com/integration/dco

perlun:
@tarcinil Sure, I'll add you - we need all the help we can get. Visit this page and you should be able to see your invitation: https://github.com/orgs/ruby-git/invitation

tarcinil:
@perlun A request has been sent to install it.

> A request to install DCO has been submitted on the @ruby-git account.

perlun:
@tarcinil Great. I approved the request now, thanks for your efforts.

PR #356, opened 2018-03-28
Tweaking yaml build to reveal git version

rvodden:
Jruby build is broken. This is an attempt to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ rvm:
   - 2.3.4
   - 2.4.3
   - 2.5.0
-  - jruby-9.1.15.0
+#  - jruby-9.1.15.0
 before_install:
   - gem install bundler
   - bundle --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,4 @@ rvm:
 before_install:
   - gem install bundler
   - bundle --version
+  - git --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,11 @@ rvm:
   - 2.3.4
   - 2.4.3
   - 2.5.0
-#  - jruby-9.1.15.0
+  - jruby-9.1.15.0
+matrix:
+  allow_failures:
+  - rvm: jruby-9.1.15.0
+  fast_finish: true
 before_install:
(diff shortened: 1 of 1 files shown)


perlun:
> Jruby build is broken. This is an attempt to fix it.

Well, "fix" is one way to put it... 😆 

I think the root cause here is that _our code is not really thread safe_, and that's why JRuby (which is much more concurrent than the evil-GIL-threaded MRI) triggers these errors.

I think we should work on fixing this; I don't think disabling CI on JRuby is the right way forward. (I looked briefly at the code in `base.rb` right now, and couldn't find any immediate non-threadsafe part.)

The error:

```
Error: test_git_init_bare(TestThreadSafety): Git::GitExecuteError: git '--git-dir=/tmp/d20180328-6779-88i76v/.git' init '--bare'  2>&1:fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>)
```

...indicates that something goes wrong; it _does_ specify the `--git-dir` but git still complains about this.

I read [this SO thread](https://stackoverflow.com/questions/25172470/git-error-fatal-git-work-tree-or-work-tree-directory-not-allowed-without) but I couldn't see the `GIT_WORK_TREE` variable being set on the Travis machines.

rvodden:
I completely agree that we need to fix Jruby, but we also need to get the
huge backlog of PRs merged in, and a temperament master build really won't
help. I suggest we take it out of the build on master, and spin up a branch
along side which automatically merges master, but has jruby enabled, so
that we can see what's going on.

perlun:
> I completely agree that we need to fix Jruby, but we also need to get the huge backlog of PRs merged in, and a temperament master build really won't help.

That I can agree with.

> I suggest we take it out of the build on master, and spin up a branch along side which automatically merges master, but has jruby enabled, so that we can see what's going on.

The easiest way for now is probably to allow JRuby to fail: https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

If we do it like that, I think I can live with it for the time being. Thanks for the effort btw!

PR #357, opened 2018-03-29
1st

vnd96:


perlun:
@VuNgocDiep96 What is the scope of this PR? Is it a mistake?

mtancoigne:
@perlun Hi ! If you look at @vnd96's profile, you'll see that this PR seems to be part of the github first steps... So I guess this PR can be safely ignored and closed...

Issue #361, opened 2018-04-19
allow fetch to receive a `ref` option


taquitos:
It would be 💯 if we could allow fetch to receive a `ref` option
One example of why it's useful:
When interacting with GitHub pull requests, you need to specify a ref: 
`git fetch origin pull/20/head`
This fetches a branch associated with a PR

taquitos:
It's been a busy few months, and I've forked the repo so we can continue moving forward. I'll try to update this soon

taquitos:
Merged pr

PR #362, opened 2018-04-19
Allow fetch operation to receive a `ref` param

taquitos:
addresses https://github.com/ruby-git/ruby-git/issues/361

```
Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the
    best of my knowledge, is covered under an appropriate open
    source license and I have the right under that license to   
    submit that work with modifications, whether created in whole
    or in part by me, under the same open source license (unless
    I am permitted to submit under a different license), as
    Indicated in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including
    all personal information I submit with it, including my
    sign-off) is maintained indefinitely and may be redistributed
    consistent with this project or the open source license(s)
    involved.
```


lib/git/base.rb
```diff
@@ -313,8 +313,8 @@ def checkout_file(version, file)
 
     # fetches changes from a remote branch - this does not modify the working directory,
     # it just gets the changes from the remote if there are any
-    def fetch(remote = 'origin', opts={})
-      self.lib.fetch(remote, opts)
+    def fetch(remote = 'origin', ref = nil, opts={})
```
kylemacey: This could potentially break implementations currently using `#fetch`. For example,

```ruby
g.fetch(origin, option: "something")
```

Will set `remote` to `origin` and `ref` to `{ option: "something"}`

Maybe we can utilize it inside of `opts`?

taquitos: Good point, will look into this.

tests/units/test_remotes.rb
```diff
@@ -122,6 +122,34 @@ def test_fetch
...
+        loc.commit('master commit 1')
+        first_commit = loc.log.first.sha
+        
+        new_file('test-file2', 'gonnaCommitYouToo')
+        loc.add
+        loc.commit('master commit 2')
+        second_commit = loc.log.first.sha
+
+        # Make sure fetch message only has the first commit when we fetch the first commit
+        assert(loc.fetch('origin', {:ref => first_commit}).include?(first_commit))
```
taquitos: Any thought on how to make this a little more robust?

kylemacey: I think it might be more pertinent to test that the `#command` method receives the expected argument strings, rather than testing the functionality of git from this high up.

Perhaps something like

```ruby
assert_send(loc, :command, ["fetch", "origin", first_commit])
```

Some other considerations:

Can these refs be an array? What if I want to do this: (from [git fetch's docs](https://git-scm.com/docs/git-fetch#_examples))

```
git fetch origin +pu:pu maint:tmp
```

This *could* be a future improvement.

I'm a little skeptical about using the name `first_commit`, when it's actually just the sha. It might be more clear to the reader if you call it `first_commit_sha`.

taquitos: Yeah, that's a bad name, I thought I'd use more than the sha, then I forgot to update the variable name.
I'll look into testing the `#command` side of things, I wasn't super familiar with how best to accomplish that, so your example is really helpful
Also, refs could be an array, but I didn't take that into account. I'll see if that's easy to support, which I bet it is. If it's too annoying to add, I'll at least add a comment about the limitation.

Thanks for reviewing!!!

taquitos: Ok, so adding `assert_send([loc.lib, :command, "fetch", ["origin", first_commit_sha]])` doesn't actually help test, as it passes as long as there isn't an error raised.
I think testing it from a high-level is probably #goodEnough

Normally I would just use `expect(loc.lib).to receive(...)` if I were using rspec, but it seems like UnitTest is just not very featured when it comes to mocking.

taquitos:
Last commit folded my last comment:

Ok, so adding `assert_send([loc.lib, :command, "fetch", ["origin", first_commit_sha]])` doesn't actually help test, as it passes as long as there isn't an error raised.
I think testing it from a high-level is probably #goodEnough

Normally I would just use `expect(loc.lib).to receive(...)` if I were using rspec, but it seems like `UnitTest` is just not very featured when it comes to mocking.

kylemacey:
> Ok, so adding assert_send([loc.lib, :command, "fetch", ["origin", first_commit_sha]]) doesn't actually help test, as it passes as long as there isn't an error raised.

@taquitos sorry that was 100% my bad. I actually didn't know the actual method to use (I'm also a bit reliant on rspec for mocking) and spit out the closest thing I could find while Googling. Maybe it's something we can improve later.

kylemacey:
@perlun @tarcinil want to give this a final :+1:?

lib/git/lib.rb
```diff
@@ -738,6 +738,7 @@ def tag(name, *opts)
 
     def fetch(remote, opts)
       arr_opts = [remote]
+      arr_opts << opts[:ref] unless opts[:ref].nil?
       arr_opts << '--tags' if opts[:t] || opts[:tags]
       arr_opts << '--prune' if opts[:p] || opts[:prune]
```
perlun: I don't like us mixing `unless foo.nil?` and `if foo` syntax like this.

Are there any strong reasons why we couldn't rephrase this as:

```ruby
arr_opts << opts[:ref] if opts[:ref]
```

The only semantic difference I can think of with these two forms is that the previous will let you pass in `false` as a ref, but... that doesn't really make much sense anyway.

taquitos: No problem! I like the use of `unless` to express the idea of `We're going to add this, unless this thing is nil`. Of course your suggestion is totally reasonable: `We're going to add this if there is something to add`. I updated the pr for the style changes 👍

taquitos:
@perlun ping 😄

lib/git/lib.rb
perlun: Thanks @taquitos. I agree that the unless pattern can make the code more readable, I just want things to be uniform.

PR #368, opened 2018-06-04
Do not always chomp output

### LINKED ISSUE #347: Git commands should not all be chomped by default
Not all git subcommand output should be chomped by default. 

For example, `git show` should never have its results chomped because when it does it does actually return different content AND a different MD5 Sum hash than what is actually stored in the git repository.

This is incredibly important for automation and config management systems that rely on the results of a `git show` to determine if file contents on disk have changed or not. Even though this is potentially an outlier use case, the fact remains that the content of a file pulled from a git commit should _NEVER_ be munged, chomped, or modified in any other way that will return different content than what is actually stored in the git commit.

smortex:
This bunch of commits does some refactoring and adds a way to optionally disable chomping the output of commands.

This fixes #347 in respect to the `#show` method.  Other methods may need similar adjustment.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -166,17 +166,17 @@ module Git
     end
 
     def object_type(sha)
-      command('cat-file', ['-t', sha])
+      command('cat-file', '-t', sha)
     end
 
     def object_size(sha)
-      command('cat-file', ['-s', sha]).to_i
+      command('cat-file', '-s', sha).to_i
     end
 
     # returns useful array of raw commit object data
     def commit_data(sha)
       sha = sha.to_s
-      cdata = command_lines('cat-file', ['commit', sha])
+      cdata = command_lines('cat-file', 'commit', sha)
       process_commit_data(cdata, sha, 0)
     end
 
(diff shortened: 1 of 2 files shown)


lib/git/lib.rb
```diff
@@ -949,7 +949,18 @@ def with_custom_env_variables(&block)
...
+      command_opts = { chomp: true, redirect: '' }
+      if opts.last.is_a?(Hash)
+        command_opts.merge!(opts.pop)
+      end
+      command_opts.keys.each do |k|
+        raise ArgumentError.new("Unsupported option: #{k}") unless [:chomp, :redirect].include?(k)
+      end
+
+      default_command_opts = { redirect: '' }
+      command_opts = default_command_opts.merge(command_opts)
```
jcouball: I think this is redundant because of line 953.

jcouball (changes requested):
Looks like this PR is behind master.  If you are willing to get it caught up, I’ll take another look and if everything works, merge it.

smortex:
@jcouball I rebased these changes on top of master.

jcouball (approved):
I prefer the keyword args for `chdir` and `chomp`.

Since it is a wide ranging change, I do have a concern that it will invalidate every open PR.  They will likely have to rebase.

PR #276, opened 2016-02-16
Fix the encoding issue - 'split': invalid byte sequence in UTF-8 (up-to-date!!)

davidvartanian:
Fix the encoding issue - 'split': invalid byte sequence in UTF-8 (Argument Error) #190 - from up-to-date master

PR #369, opened 2018-06-11
process_full_diff: Don't reencode diffs which are already UTF-8

tbrand:
## Abstract

I'm heavy user of https://github.com/danger/danger and it internally uses this gem.
Danger frequently refers the `git diff`.

I noticed when `git diff` contains multi bytes text, the `full_diff_utf8_encoded` includes invalid chars like below.

I saw https://github.com/ruby-git/ruby-git/pull/276 and I could understand the background of the code.
The `command_lines` method had been fixed for it (https://github.com/ruby-git/ruby-git/commit/fe680f8c64d4b78280d541fe58e30ed1f3b9c6e0) but the `process_full_diff` had not.

## Changes

Do not force to encode diff patch when the encoding type of it is UTF-8.

### DIFF
--- a/lib/git/diff.rb
+++ b/lib/git/diff.rb
@@ -127,10 +127,11 @@ module Git
         }
         final = {}
         current_file = nil
-        full_diff_utf8_encoded = @full_diff.encode("UTF-8", "binary", {
-			:invalid => :replace,
-			:undef => :replace
-		})
+        if @full_diff.encoding.name != "UTF-8"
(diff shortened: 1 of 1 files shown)


tbrand:
ping

tbrand:
Thanks for merging this. When the patch will be released?

perlun:
@tbrand Here we go: https://github.com/ruby-git/ruby-git/releases/tag/v1.5.0 🎉

Issue #358, opened 2018-03-31
JRuby build is intermittently failing


rvodden:
From @perlun 

> I think the root cause here is that our code is not really thread safe, and that's why JRuby (which is much more concurrent than the evil-GIL-threaded MRI) triggers these errors.
> [..]
> The error:
> 
> Error: test_git_init_bare(TestThreadSafety): Git::GitExecuteError: git '--git-dir=/tmp/d20180328-6779-88i76v/.git' init '--bare'  2>&1:fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>)
> ...indicates that something goes wrong; it does specify the --git-dir but git still complains about this.
> 
> I read[ this SO thread](https://stackoverflow.com/questions/25172470/git-error-fatal-git-work-tree-or-work-tree-directory-not-allowed-without) but I couldn't see the GIT_WORK_TREE variable being set on the Travis machines.

jcouball:
I believe this was fixed in #403 and released in [v1.6.0.pre1](https://github.com/ruby-git/ruby-git/releases/tag/v1.6.0.pre1)

PR #372, opened 2018-06-25
Isolate Dir.chdir to a new process, or mutex

### LINKED ISSUE #355: Using one repo per thread results in errors
Lots of warning: conflicting chdir during another chdir block when I try to do anything with individual git objects.

I have multiple repos that I've opened using `Git.open()` and I can only operate on one at a time or else I get warnings

taquitos:
Instead of using chdir blocks, we should use absolute paths so that we can enable threaded usage of the library

If `Process.fork()` is available, we'll use that for `git.chdir` actions, otherwise, we'll use a global mutex to prevent threads from calling `git.chdir` before the previous execution has completed.

Fixes #355

Note: if you do async work in `git.chdir` all bets are off.

Signed-off-by: Joshua Liebowitz


lib/git/lib.rb
```diff
@@ -442,11 +442,7 @@ def config_get(name)
         command('config', ['--get', name])
       end
 
-      if @git_dir
-        Dir.chdir(@git_dir, &do_get)
-      else
-        build_list.call
```
taquitos: this is a typo in the original code, submitting a pr for this

perlun:
@taquitos Much appreciated! 👏 Let me know once you have this under control.

Once this is sorted out, we'll probably get #358 fixed for free.

taquitos:
I am now testing with the latest change. Will update after we have it running for a bit.

lib/git/lib.rb
```diff
@@ -442,11 +442,7 @@ def config_get(name)
         command('config', ['--get', name])
       end
 
-      if @git_dir
-        Dir.chdir(@git_dir, &do_get)
-      else
-        do_get.call
-      end
+      do_get.call(@git_dir)
```
taquitos: not 💯 this will work, but tests pass. Seems like we might need to pass the `path` to the `config`  command.

lib/git/lib.rb
```diff
@@ -458,11 +454,7 @@ def config_list
         parse_config_list command_lines('config', ['--list'])
       end
 
-      if @git_dir
-        Dir.chdir(@git_dir, &build_list)
-      else
-        build_list.call
-      end
+      build_list.call(@git_dir)
```
taquitos: same here, I think we need to utilize the `path` variable in the `config` command if it exists

lib/git/lib.rb
```diff
@@ -8,6 +8,7 @@ class GitExecuteError < StandardError
   class Lib
 
     @@semaphore = Mutex.new
+    @@config_semaphore = Mutex.new
```
taquitos, line 11: calls to config also should be protected, but it might be heavy handed to `fork`

taquitos:
@perlun I'd love to get your thoughts. The changes don't make everything thread-safe, but it takes care of a lot of issues. There are a few places where paths can't be passed in `$git config` is one of those, so we must `Dir.chdir` 🙄when we're dealing with local config.

perlun:
@taquitos Thanks, will try to get this reviewed ASAP. Two minor things:

1. Could you rebase/merge in master into the branch?
2. Could you add the DCO?

Thanks in advance.

taquitos:
@perlun when you get a moment 👍

lib/git/base.rb
```diff
@@ -3,6 +3,12 @@
 module Git
   
   class Base
+    # Adding a mutex to the class because each repo should be sharing the same mutex 
+    # in case we need to Dir.chdir and we don't have fork() support to isolate that
+    class << self
+      attr_accessor :chdir_semaphore
+    end
+    Git::Base.chdir_semaphore = Mutex.new
```
perlun: This works perfectly fine, but maybe the approach below would be even better. It shouldn't really be an accessor but more like a reader since it's an immutable data structure that shouldn't be tampered with elsewhere, right?

```ruby
class << self
  def chdir_semaphore
    @chdir_semaphore ||= Mutex.new
  end
end
```

lib/git/base.rb
```diff
@@ -92,18 +98,37 @@ def initialize(options = {})
...
+      end
+
+      if Process.respond_to?(:fork)
+        # Forking this process so that we can be threadsafe
+        pid = Process.fork do
+          chdir_block.call
+        end
+        Process.wait(pid)
+      else
+        # Windows and NetBSD 4 don't support fork()
```
perlun: Interesting anecdote, I wasn't aware of NetBSD 4 here. But OTOH, it's past EOL already. Maybe we can drop that reference and instead mention JRuby (which is actively used by many), since it also has problems with `fork()` on the JVM.

lib/git/base.rb
```diff
@@ -144,9 +169,7 @@ def repo
     
     # returns the repository size in bytes
     def repo_size
-      Dir.chdir(repo.path) do
-        return `du -s`.chomp.split.first.to_i
-      end
+      return `du -s #{repo.path}`.chomp.split.first.to_i
```
perlun, line 172: Very good fix, absolutely no reason to `chdir` in this case. :man_facepalming:

lib/git/lib.rb
```diff
@@ -310,7 +311,7 @@ def branches_all
     def list_files(ref_dir)
       dir = File.join(@git_dir, 'refs', ref_dir)
       files = []
-      Dir.chdir(dir) { files = Dir.glob('**/*').select { |f| File.file?(f) } } rescue nil
+      files = Dir.glob(File.join(dir, '**/*')).select { |f| File.file?(f) }
```
perlun: I wonder if the `rescue nil` removal can be dangerous? Unfortunately not aware of its root cause of being there though.

lib/git/base.rb
taquitos: True dat

lib/git/base.rb
taquitos: I think you're right, but are you sure that's threadsafe? If we have 2 threads that call `chdir_semaphore` before it is set, is it possible they both end up calling `Mutex.new`?

lib/git/lib.rb
taquitos: Yeah, I was thinking about that too. I wasn't convinced it was needed, but it's probably safer since passing `nil` into `File.file?` or `Dir.glob` can raise an error

lib/git/base.rb
perlun: True. So maybe something like this then:

```ruby
@chdir_semaphore = Mutex.new

class << self
  attr_reader :chdir_semaphore
end
```

(if I remember the context right - the `@chdir_semaphore call in normal class context will assign it to the class-level variable. Please try this out before trusting me blindly on this.)

taquitos: @perlun I tested with your suggestion, it works 👍

perlun:
@taquitos Is this reviewable now?

taquitos:
Ugh. Making new PR

PR #374, opened 2018-07-02
commit with custom author date

mgarciaisaia:
Honor `git commit --date=...` flag.

Re-submitting the stalled #315

### DIFF
--- a/tests/units/test_lib.rb
+++ b/tests/units/test_lib.rb
@@ -21,7 +21,20 @@ class TestLib < Test::Unit::TestCase
     assert_equal("test\n", data['message'])
     assert_equal(["546bec6f8872efa41d5d97a369f669165ecda0de"], data['parent'])
   end
-  
+
+  def test_commit_with_date
+    create_file("#{@wdir}/test_file_1", 'content tets_file_1')
+    @lib.add('test_file_1')
+
+    author_date = Time.new(2016, 8, 3, 17, 37, 0, "-03:00")
+
+    @lib.commit('commit with date', date: author_date.strftime('%Y-%m-%dT%H:%M:%S %z'))
+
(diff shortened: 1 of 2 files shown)


tests/units/test_commit.rb
```diff
@@ -0,0 +1,53 @@
...
+
+require File.dirname(__FILE__) + '/../test_helper'
+
+class TestCommit < Test::Unit::TestCase
+
+  def setup
+    set_file_paths
+  end
+
+  def test_add
```
tarcinil: This should be in `tests/units/test_lib.rb` to ensure that things in `Git::Lib` are tested where they are expected.

The test name makes more sense as `#test_commit_with_date` in that file.

mgarciaisaia:
Thanks for the feedback, @tarcinil. It seems the project moved really forward since I've submitted the original PR - awesome! 🎉 

I've updated the tests to fit the ones that previously existed. It even seems more clear to me, now.

Let me know what you think 👍

tests/units/test_lib.rb
```diff
@@ -21,7 +21,22 @@ def test_commit_data
...
   end
-  
+
+  def test_commit_with_date
+    create_file("#{@wdir}/test_file_1", 'content tets_file_1')
+    @lib.add('test_file_1')
+
+    author_date = Time.new(2016, 8, 3, 17, 37, 0, "-03:00")
+
+    previous = @lib.commit_data('HEAD')
```
tarcinil: What does this provide?  I do not see it being used anywhere.

mgarciaisaia:
Good catch! It was a leftover from a previous version of the test.

mgarciaisaia:
Sorry - hoped you could rebase it from here. There it is! 👍

PR #377, opened 2018-07-26
Add support for unshallow

singpolyma:
Signed-off-by: Stephen Paul Weber <singpolyma@singpolyma.net>

Allow passing the `--unshallow` option

### DIFF
--- a/tests/units/test_lib.rb
+++ b/tests/units/test_lib.rb
@@ -13,7 +13,16 @@ class TestLib < Test::Unit::TestCase
     set_file_paths
     @lib = Git.open(@wdir).lib
   end
-  
+
+  def test_fetch_unshallow
+    in_temp_dir do |dir|
+      git = Git.clone("file://#{@wdir}", "shallow", path: dir, depth: 1).lib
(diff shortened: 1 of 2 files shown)


tarcinil:
@singpolyma 

How can this be tested?  Can you create a shallow repo, and use this to make it an unshallow?

singpolyma:
@tarcinil test added

tests/units/test_lib.rb
```diff
@@ -13,7 +14,16 @@ def setup
     set_file_paths
     @lib = Git.open(@wdir).lib
   end
-  
+
+  def test_fetch_unshallow
+    Dir.mktmpdir("ruby-git-shallow") do |dir|
```
tarcinil: You do not need to include/require that.

We have `in_temp_dir` inside our helper that can do this work for you.  You can see an example in https://github.com/ruby-git/ruby-git/blob/master/tests/units/test_stashes.rb#L11

singpolyma: updated

PR #381, opened 2018-08-10
Bug with a .gitignore

shuklineg:
When you use directories or masks in a .gitignore, the files ignored by git fall into the list of untracked files

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -428,8 +428,11 @@ module Git
     def ignored_files
       command_lines('ls-files', ['--others', '-i', '--exclude-standard'])
     end
-
-
+    
+    def untracked_files
+      command_lines('ls-files', ['--others','--exclude-standard'])
+    end
+    
     def config_remote(name)
       hsh = {}
       config_list.each do |key, value|
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -170,14 +170,11 @@ module Git
     end
 
     def fetch_untracked
-      ignore = @base.lib.ignored_files
+      untracked = @base.lib.untracked_files
 
       Dir.chdir(@base.dir.path) do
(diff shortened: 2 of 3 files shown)


tarcinil:
@shuklineg Thank you for your contribution.  

Please add DCO to your commits and add tests that show that your changes are working as expected for the use-case that you presented.  

My idea would to be create a test that creates a new repo, adds a few files (including a .gitignore), the gitignore should ignore one of the files added, and you should test that that file is not in the untracked output.

cpb:
@tarcinil this fixes it for me. @EPecherkin can you review this please?

detlefw:
This fix does not work for me, if the current working directory of the program is not the base directory of the repository. If they are both the same it works as expected. For me the fix is, to change the Git::Lib#command method. Setting the current working directory for git with
```ruby
      global_opts << "--git-dir=#{@git_dir}" if !@git_dir.nil?
      global_opts << "--work-tree=#{@git_work_dir}" if !@git_work_dir.nil?
```
in the command method does not work as expected. A working fix for me is to use git's -C parameter as:
```ruby
      global_opts << ['-C', @git_work_dir] if @git_work_dir
```

jcouball:
@detlefw I was unable to reproduce what you reported above.  Can you please submit a way to reproduce the error?  In the form of a failing unit test would be idea, but any reproduction would be appreciated.

> This fix does not work for me, if the current working directory of the program is not the base directory of the repository. If they are both the same it works as expected.
> ...

If you submit something, I'll take a look and hopefully move this along.

pcarn:
When I have 69,000 files in my `node_modules` directory, which is in `.gitignore`, it takes about 25 seconds to run `Git.open('.').status`.
Which these changes, it reduces the time to 0.5 seconds. 
This would greatly help.

SolaWing:
what's the status of this patch? why it's not got merged?

PR #385, opened 2018-08-25
fixing issues with a HEREDOC entry in git.status

### LINKED ISSUE #384: repo.status.pretty fails
### Subject of the issue
`repo.status.pretty` gives the following error:
`gems/git-1.4.0/lib/git/status.rb:114:in 'pretty_file': undefined method 'strip_heredoc' for #<String:0x00007ffee88a0c20> (NoMethodError)`

It appears that `.strip_heredoc` method (used [here](https://github.com/ruby-git/ruby-git/blob/master/lib/git/status.rb#L107)) is a rails method not ruby - https://apidock.com/rails/String/strip_heredoc

One solution to fix this might be Ruby 2.3 native support for this with the `<<~` operator. However, this has the obvious downside of limiting possible use of this library.
Or possibly to implement a method to achieve the same goal, e.g. copy paste this - https://github.com/rails/rails/blob/b422cda2ebfff4032f4c18271e96ad329c413dcc/activesupport/lib/active_support/core_ext/string/strip.rb

If you'd like I can put together a PR with either solution if you'd like?

### Your environment
```
$ git --version
git version 2.14.3 (Apple Git-98)
$ ruby --version
ruby 2.5.0p0 (2017-12-25) [x86_64-darwin16]
```

### Steps to reproduce
In a blank project with just this dependency this code will fail:
```ruby
repo = Git.clone("git@github.com:#{destination_repository}.git", destination_repository, path: path)
# make some changes to the repo
puts repo.status.pretty
```
### Expected behaviour
The method call would return a pretty string to describe the repo status
### Actual behaviour
An error is raised

tarcinil:

### Description
fix #384

Removes `#strip_heredoc

This change **breaks** Ruby 2.2 or less and the `<<~` only exists in Ruby 2.3+.  We are only testing Ruby 2.3 and later in Travis CI and I would argue that we should only be supporting those in the newer versions of the gem.

Fixing branch for a message being used interpreted as an arg list and not as a decomposed argument
```
Code/ruby-git/lib/git/branch.rb:40: warning: parentheses after method name is interpreted as an argument list, not a decomposed argument
```


tarcinil:
@jcouball - I don't have an argument either way, I would be fine with a MINOR.  The change in the gemspec should prevent`bundle` from allowing users to update if they don't have Ruby 2.3+.  The overall API of the method is staying the same, and it is being removing a dependency on a rails only function (`#strip_heredoc`) which we do not include.

jcouball:
Let's let this PR go with just a patch version bump.  By semver definition, that is all that is needed.

stephengroat:
any status on this?

PR #391, opened 2018-10-09
Fix for Bad character encoding.

c-f:
Signed-off-by: CF <cf@data-co.de>

### Description
When files contain bad characters (e.g. \255) and commits are compared an ArgumentError is thrown. 
This PR changes handling bad characters in the else statement. 

In addition to PR #369 when converting from UTF-8 to UTF-8 no valid high acii characters are changed.
#### Test script
```ruby
#!/bin/env ruby
# encoding: UTF-8
# Purpose: Encoding testing

# --[Test strings]--
test_strs = [
    "红宝石",
    "\255",
    "红宝石\255",
]

puts "Encoding Name is UTF-8:"
puts ["String","UTF-8","Valid?"].join("\t\t")
test_strs.each do |str| 
    puts [
        str,
        str.encoding.name == "UTF-8",
        str.valid_encoding?
    ].join("\t\t")
end
puts "------------------------"

puts "Get Encoding (fixed):"
puts ["String","Encoded","Split"].join("\t\t")
test_strs.each do |str|
    if str.encoding.name != "UTF-8"
        encoded = str.encode("UTF-8", "binary", { :invalid => :replace, :undef => :replace })
    else
        encoded = str.encode("UTF-8", "UTF-8",  { :invalid => :replace, :undef => :replace })
    end  
    puts [
        str,
        encoded,
        encoded.split("").join(""),
    ].join("\t\t")
end
puts "------------------------"

puts "Get Encoding (current):"
test_strs.each do |str|
    puts str.split("").join("")
end
puts "------------------------"
```

#### Output 
```
Encoding Name is UTF-8:
String		UTF-8		Valid?
红宝石		true		true
�		true		false
红宝石�		true		false
------------------------
Get Encoding (fixed):
String		Encoded		Split
红宝石		红宝石		红宝石
�		�		�
红宝石�		红宝石�		红宝石�
------------------------
Get Encoding (current):
String		Encoded		Split
红宝石		红宝石		红宝石
Traceback (most recent call last):
	3: from test_utf8.rb:42:in `<main>'
	2: from test_utf8.rb:42:in `each'
	1: from test_utf8.rb:51:in `block in <main>'
test_utf8.rb:51:in `split': invalid byte sequence in UTF-8 (ArgumentError)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ And here are the operations that will need to write to your git repository.
      g.merge(g.branch('master'))
      g.merge([branch1, branch2])
 
+     g.merge_base('branch1', 'branch2')
+
      r = g.add_remote(name, uri)  # Git::Remote
      r = g.add_remote(name, Git::Base)  # Git::Remote
 
--- a/lib/git/base/factory.rb
+++ b/lib/git/base/factory.rb
@@ -68,6 +68,14 @@ module Git
         Git::Object.new(self, tag_name, 'tag', true)
       end
 
+      # Find as good common ancestors as possible for a merge
+      # example: g.merge_base('master', 'some_branch', 'some_sha', octopus: true)
+      # returns Array<Git::Object::Commit>
+      def merge_base(*args)
+        shas = self.lib.merge_base(*args)
+        shas.map { |sha| gcommit(sha) }
+      end
+
     end
 
   end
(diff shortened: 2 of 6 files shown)


c-f:
Sorry - messing with up with DCO -.-

PR #398, opened 2019-01-21
Allow creating new config entries

mpapis:

### Description

[Git 2.1.2 introduced setting values even if they did not previously exist.](https://github.com/git/git/blob/master/Documentation/RelNotes/2.1.2.txt#L16-L17)
The git-ruby library [keeps compatibility](https://github.com/ruby-git/ruby-git/blob/master/lib/git.rb#L25) with Git [1.6](https://github.com/ruby-git/ruby-git/blob/master/lib/git/lib.rb#L887), this fixes the
problem of older versions not being able to create new values.

#### Other options

- bump the minimal required git version to `2.1.2`

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -502,10 +502,14 @@ module Git
 
     def config_set(name, value)
       command('config', [name, value])
+    rescue Git::GitExecuteError
+      command('config', ['--add', name, value])
     end
 
     def global_config_set(name, value)
       command('config', ['--global', name, value], false)
+    rescue Git::GitExecuteError
+      command('config', ['--global', '--add', name, value], false)
     end
 
     # updates the repository index using the working directory content


mpapis:
tests are passing, the code is rebased, anything else I should do?

mpapis:
still waiting for the review

mpapis:
still waiting for the review

mpapis:
still waiting for the review

mpapis:
still waiting for the review

jcouball:
I suspect that requiring 2.1.2 as the minimum git version might be burdensome to users.  According to [this page](https://en.wikipedia.org/wiki/Git), 2.1.2 was released in 2017.  I am interested in hearing what others think.

Looks like the implementation is to try the `git config` command without the `--add` switch first and then try using the `--add` switch if that fails.  While it works, it seems a bit sloppy since the initial command could have failed for other reasons.  That said, I can offer no better alternative at the moment.

Some things to consider as we evolve ruby-git further:
* Maybe the library should discover what version of git is being used and act more purposefully on what features are supported by the git client and which are not.
* It would be nice to be able to test with multiple versions of git so ruby-git's interaction with each version can be validated.

jcouball:
Have you thought about adding a test for the situation covered by the change?  Is it possible with the existing state of the testing framework?

mpapis:
I think it was to hard, the two other PR's from this time got tests (#399, #400), I'm moved already to other company, keeping the PR's alive as a courtesy to other OSS users.

jcouball:
I am closing this PR to have a broader discussion about how `git config` should work in this gem in the next major release in #656.

PR #400, opened 2019-01-21
Refactor git config

mpapis:
### Your checklist for this pull request
🚨Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

- [x] Ensure all commits include DCO sign-off.
- [x] Ensure that your contributions pass unit testing.
- [x] Ensure that your contributions contain documentation if applicable.

### Description

The Git module has config and global_config methods that are available only when Git module is included. Refactored the code to remove duplication and to make it more explicit we are managing `git config` not the library setup.

PR #403, opened 2019-01-30
Do not allow changes to ENV to leak from test to test

jcouball:

### Description
Currently, the TestThreadSafety test case occasionally fails when executed as part of the build with this error:

```
#<Thread:0x00007fd08ea136f8@/Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/tests/units/test_thread_safty.rb:19 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	4: from /Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/tests/units/test_thread_safty.rb:20:in `block (2 levels) in test_git_init_bare'
	3: from /Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/lib/git.rb:140:in `init'
	2: from /Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/lib/git/base.rb:67:in `init'
	1: from /Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/lib/git/lib.rb:40:in `init'
/Users/couballj/SynologyDrive/Documents/Projects/ruby-git-master-2/lib/git/lib.rb:983:in `command': git '--git-dir=/var/folders/xv/2j9g0mgs5vgcdrl0s9pn9l_40000gn/T/d20190129-61742-zetuf4/.git' init '--bare'  2>&1:fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>) (Git::GitExecuteError)
```

the important part is:

```
git '--git-dir=/var/folders/xv/2j9g0mgs5vgcdrl0s9pn9l_40000gn/T/d20190129-61742-zetuf4/.git' init '--bare'  2>&1:fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>)
```

It can be made to fail consistently by greatly increasing the thread count.  50 threads on my test machine makes if fail every time.

I found that other tests alter the ENV global variable which makes the test_thread_safety test suite fail.

This change makes it so that tests always restore ENV to its original state.

I think this is a short term fix.  A better fix would be to pass the desired environment into any method that creates a Git::Base (defaulting to a copy of ENV) and not use the ENV global through out the code.

Signed-off-by: James Couball <jcouball@yahoo.com>


jcouball:
I believe this fixes issues #319 and #358 as well as the build issue reported in PR #400

jcouball:
@perlun and @tarcinil -- this PR is ready for review.

tests/test_helper.rb
```diff
@@ -79,5 +79,19 @@ def append_file(name, contents)
...
+  #
+  def with_custom_env_variables(&block)
+    saved_env = {}
+    begin
+      Git::Lib::ENV_VARIABLE_NAMES.each { |k| saved_env[k] = ENV[k] }
+      return block.call
+    ensure
+      Git::Lib::ENV_VARIABLE_NAMES.each { |k| ENV[k] = saved_env[k] }
+    end
+  end
```
perlun, line 96: (This reminds me about my dislike for `test-unit`; with `rspec`, this could easily be done using an `around {}` block instead.)

jcouball: I prefer `rspec` for the structure it makes possible.  I suspect that others don't care for `rspec` because that structure comes with additional complexity and a steeper learning curve.

That said, I respect any project that has tests no matter what framework!

perlun: Comparing `rspec` to other test frameworks in general, like `testng` in Java, I think the fact that you can actually "group" and structure your tests with `rspec` is a big boon. It's one of the things I'm missing from no longer using it actually.

If you feel like rewriting the tests to use `rspec`, I clearly wouldn't object, but this is definitely not an obligation in any way. 😃

PR #405, opened 2019-02-28
Do not let unexpected command output encoding raise an exception

jcouball:
Signed-off-by: James Couball <jcouball@yahoo.com>

### Description

ruby-git relies on shelling out to the `git` command to interact with a `git` repository.  When the `git` command has output that is not in the default Ruby encoding (usually UTF-8), `ruby-git` may raise an exception when the string containing the command output is used.

This change makes `ruby-git` check if the command output string contains any invalid code points and if it does, attempts to guess at the actual encoding of the command output on a line-by-line basis.  Each line is converted from the actual encoding to Ruby's default encoding (usually UTF-8).  Unknown code points are replaced by a special character.

This solution contrasts with PR #404 which simply tries to replace all invalid code points with a special character using `String#scrub`.

### Background

`git diff` (and other git commands) can have output that has mixed encodings.  Not only does `git` output its own text in the default system/shell encoding, but when text from a repository file is output, it is not in any particular encoding: `git` just passes the bytes from the file to the output. These files are stored in whatever encoding the user wrote them in.

In Ruby, `popen` and `backtick` collect the output from the command they run as bytes.  They shove these bytes into the string returned.  They make no attempt to guess the *actual* encoding of the string returned.

In later versions of Ruby (maybe 1.9+), the encoding of the string returned from `popen` or `backtick` is ALWAYS set to `Encoding.default_external`.  This is usually ‘UTF-8’.  Ruby uses the value of the command `locale charmap` for the initial value of `Encoding.default_external`.  `popen` and `backtick` use this encoding even though the *actual* encoding of the command output is not known.

The bottom line is that the *actual* encoding of the string returned from `popen` or `backtick` is unknowable.  Sometimes it is a happy coincidence that the *actual* encoding matches what is retuned by `String#encoding`.  When `Encoding.default_external` is ‘UTF-8’ this coincidence happens more often than not.

I think the best that can be done is to sanitize the command output as it is read in `Git::Lib#run_command` in an attempt to “guard the gates”.  That way the rest of the code in this gem can assume the command output has a valid encoding and avoid unexpected exceptions being raised.

### What has changed?

* `git.gemspec`: added a runtime dependency on [rchardet](https://github.com/jmhodges/rchardet) to guess the encoding of command output in the case where the command output contains invalid code points.
* `lib/git/lib.rb`: change `Git::Lib#run_command` to ensure that the string that collects command line output is converted to the String encoding used by Ruby and only has valid code points for that encoding.
* `lib/git/diff.rb`: remove encoding handling since this is now handled in `Git::Lib#run_command`
* `tests/units/test_diff_non_default_encoding.rb`: Added two tests: (1) diff changes to a file that has iso-8859-7-greek encoding and (2) diff changes to a file where the encoding was changed from euc-jp-japanese to euc-kr-korean
* `tests/files/encoding/**/*`: a repository used in the tests above


benasher44:
Any chance of seeing this merged? Just ran into this, and we've resorted to forking the repo and keeping this branch up-to-date

agentk:
Getting the same issue too. It's affecting Danger from being able to get the list of changed files when one of the changed files (correctly) contains invalid UTF-8.

genepaul:
I would also like to see this fixed, as we are running into the same problem.

kielgillard:
👍 for seeing this merged sooner rather than later.

Kaspik:
@tarcinil / @perlun ? Pretty please!

Kaspik:
So what can we do about it? Also, `rchardet` has some Snyk Vulnerability reported.

perlun:
@Kaspik 

> So what can we do about it?

We can kindly ask @jcouball to see if there are other ways to achieve (reasonably) the same goal without the added `rchardet` dependency.

jcouball:
Hello @perlun and @kaspik, sorry to take so long to respond.

I haven't had much luck finding a suitable replacement for `rchardet`.  The choices are very few and not very well maintained.

Also, I haven't found a vulnerability reported for `rchardet` here:
https://snyk.io/vuln/search?q=rchardet&type=rubygems

Is there somewhere else I should check?

Assuming that there is no vulnerability, would you accept this PR as is?  I could look into making the `rchardet` dependency optional... this might be good intermediate step to making this functionality default.

What do you think?

perlun:
Sorry @jcouball, this time I'm the one who's being a bit slow here... 🙈 

> Assuming that there is no vulnerability, would you accept this PR as is? I could look into making the `rchardet` dependency optional... this might be good intermediate step to making this functionality default.

Maybe it's not the end of the world with this added dependency after all. If you rebase the MR, I'll approve it.

Kaspik:
@jcouball Hey! Are you planning to rebase this PR? :)

jcouball:
Hello @perlun and @Kaspik, I have rebased this PR.

Kaspik:
Thanks! Looks good to me, now it's up to @perlun :)

jcouball:
@perlun, can you please merge this?

perlun:
@jcouball Approved, sorry for the unnecessary delay from my part. Please merge yourself so the commit history gets more correct.

jcouball:
@perlun, what is the plan / process to create a new release?

perlun:
> @perlun, what is the plan / process to create a new release?

There is not much of a "plan" at the moment unfortunately (I'm not actually using the gem myself any more), but the process is roughly this, using the `release[remote]` Rake task below:

```shell
$ rake -T
rake build            # Build git-1.5.0.gem into the pkg directory
rake clean            # Remove any temporary products
rake clobber          # Remove any generated files
rake install          # Build and install git-1.5.0.gem into system gems
rake install:local    # Build and install git-1.5.0.gem into system gems without network access
rake release[remote]  # Create tag v1.5.0 and build and push git-1.5.0.gem to rubygems.org
rake test             # Run Unit Tests
```

Inspired by this guide: https://bundler.io/v1.12/guides/creating_gem.html#releasing-the-gem, I'd describe the steps like this:

* Bump the version number in `lib/git/version.rb` - definitely use Semantic Versioning here. Are we aiming for a 1.6 or 2.0 release here, WDYT?
* Commit this change. If you have [`git-extras`](https://github.com/tj/git-extras) installed, you can do `git release v1.6.0` or similar.
* Run the `rake release[remote]` task - this presumes you are set up as a Rubygems maintainer for this gem.
* Create a GitHub release for the new tag and add the changelog. This is typically created using [changelog-rs](https://github.com/perlun/changelog-rs) which makes the process reasonably simple, as long as the commit messages are good enough and all branches are squash-merged into master.

@jcouball, if we can settle on a version number, I can probably help out with the actual release in this case. For bonus points, you could document the steps above in a file in the repo so we remember it until next time. (which makes things simpler for everybody)

majioa:
Can't validate for now, let's release 1.6.0 then it will have seen there

PR #406, opened 2019-03-02
Add DiffFile#added_lines

Kache:

### Description
`DiffFile#added_lines` returns the source lines added and their respective line numbers.

My use case involves adding [Danger](https://github.com/danger/danger) warnings to PRs based on inserted lines.

Wanted to get some [feedback](https://github.com/ruby-git/ruby-git/issues/407) before putting more work into it (tests, etc).


jcouball:
I think the idea of this PR looks fine.  I'd like to generalize it to return both added and deleted lines which seems simple enough.

Should we introduce a new abstraction to represent a 'difference line' to encapsulate the line, line number, and maybe if the line was added or deleted?

I added other comments inline.

lib/git/diff.rb
```diff
@@ -95,6 +95,32 @@ def blob(type = :dst)
...
+          after_lines = code_lines.reject { |line| line.start_with?('-') }
+          numbered_lines = after_line_nums(range_info).zip(after_lines)
+          numbered_lines.map! { |n, line| [n, line[1..-1]] if line.start_with?('+') }.tap(&:compact!)
+        end
+      end
+
+      private
+
+      def change_hunks
+        patch.each_line.chunk_while { |_, nl| !nl.match?(range_info_regex) }.drop(1)
```
jcouball: The use of `range_info_regex` here confused me at first.  In this context, I think `hunk_header?` would make it easier to understand.  Maybe extract `nl.match?(range_info_regex)` to a method `def hunk_header?(line)`.

lib/git/diff.rb
```diff
@@ -95,6 +95,32 @@ def blob(type = :dst)
...
+      end
+
+      def after_line_nums(range_info)
+        m = range_info_regex.match(range_info)
+        starting_num, num_lines = [m[2], m[4] || 1].map(&:to_i)
+        starting_num...(starting_num + num_lines)
+      end
+
+      def range_info_regex
+        @range_info_regex ||= /^@@ -\d+(,\d+)? \+(\d+)(,(\d+))? @@/
```
jcouball: No need to memoize the regex here.  The regex is treated as a constant:

```Ruby
irb
2.6.1 :001 > class Foo
2.6.1 :002?>   def my_regex
2.6.1 :003?>     /^test/
2.6.1 :004?>     end
2.6.1 :005?>   end
 => :my_regex
2.6.1 :006 > foo1 = Foo.new
 => #<Foo:0x00007f997e922f10>
2.6.1 :007 > foo1.my_regex
 => /^test/
2.6.1 :009 > foo2 = Foo.new
 => #<Foo:0x00007f997e9326b8>
2.6.1 :010 > foo1.my_regex.object_id
 => 70148615101060
2.6.1 :011 > foo1.my_regex.object_id
 => 70148615101060
2.6.1 :012 > foo2.my_regex.object_id
 => 70148615101060
2.6.1 :013 >
```

Kache:
Sure, I had considered that too, although I thought it had some further-reaching consequences I was avoiding:

 1. Ideally in the long run, the diff-parsing of my private `DiffFile` methods and `process_full_diff` would be consolidated, as they conceptually share a single responsibility.
 2. It looks like there was an intention to refactor `process_full_diff`?
 3. If I were to implement `DiffLine` without touching `process_full_diff`, I'd likely be adding a decent amount of tech debt, and every line in the diff would also end up getting unnecessarily parsed twice.

Your thoughts? Btw, I don't see any comments inline.

lib/git/diff.rb
jcouball: @Kache sorry, I didn't submit the conversation entries I had made -- they are there now.  

I agree about consolidating the parsing is probably the way to go.  Would be nice if people who did not need access to individual lines of diff did not have to pay the price of parsing them.

jcouball:
@Kache sorry, I didn't submit the conversation entries I had made -- they are there now.

I agree about consolidating the parsing is probably the way to go. Would be nice if people who did not need access to individual lines of diff did not have to pay the price of parsing them.

lib/git/diff.rb
Kache: whoa, cool! This doesn't seem to happen for other objects (Hash, Array), even when I freeze them. Is this specific to `Regexp`, like `Symbol`? It's not even the exact same as symbol, i.e. `/foo/.equal?(/foo/) # false`. Do you know where I can find documentation?

Kache:
Hi @jcouball, I've updated this PR in accordance everything's that's been discussed so far.

The most sizable part is the parsing refactor, which I've grouped together with a non-ff merge:



Let me know if you'd prefer it as a separate PR that merges first. Note that it's not a perfect/pure refactor, there are a few minor changes that may warrant evaluation.

In any case, I've organized the changes as atomic commits for ease of review. All tests should pass at every commit.

Kache:
Update: I'm going to try an alternative -- rather than having both `#added_lines` and `#deleted_lines`, to have a single `#each_diff_line` (aliased by `#each`) that returns both as interleaved `DiffLine` objects.

Issue #23, opened 2011-05-20
ruby-git status and command line git status differ 


josh803316:
I'm attempting to check for git changes in a current directory after I re-write a file.  For some reason @git.status.changed always shows that the file has been changed, even though command line git status shows no diff, as it should.  Is there a way around this.  I'm attempting to decide whether to update a field in the file if it has been changed before pushing to git.

```
    # Check the if the contents of a directory in git
    def dir_has_changes? dir
        intersection = dir.split('/') - @top_level.split('/')
        intersection = intersection.join('/')
        has_changes = false
        @git.status.changed.each do |file|
            if file[0] =~ /#{intersection}/
                has_changes = true
            end
        end
        return has_changes
    end
```

chilicheech:
This is affecting me as well. Why can't ruby-git `status` call `git status` instead of `git diff-files`? This implementation is confusing because it differs from `git`'s implementation.

jon-rtr:
I'm running into the same behavior as listed above, so I would say this issue should be reopened.

> I'm having this issue as well.
> 
> EDIT
> 
> Curiously, shelling out to git status from ruby before, e.g. @git.status seems to result in the latter reporting status accurately.

I'm using `ruby-git 1.5.0` and `git version 2.15.2 (Apple Git-101.1)`.

jon-rtr:
Running git diff-files on one of my repos, all of the files look like this:
```
$ git diff-files | head -n2
:100644 100644 51e14f30f818cde956880833f3c9a5822b54ad03 0000000000000000000000000000000000000000 M	.gitignore
:100644 100644 93deb154ae39ea2ae76c9a70ddf51be9dcbd8f4a 0000000000000000000000000000000000000000 M	README.md
```

tarcinil:
@jon-rtr 

Thank you for your report.  As the issue is now seven years old and we can't be sure how the issue has changed over the many years.  Please create a new issue will all relevant data.

PR #409, opened 2019-03-14
Update index before fetching modified files

ggilder:
According to the [git-diff-index docs][1], git commands like diff-index
and diff-files typically don't look at the actual contents of files in
the working tree. This means that these commands can be easily fooled by
operations that don't change the content of a file such as deleting and
recreating it, or even running `touch` on a file.

The docs then say that one may update the index to make it be in sync
with the actual contents of the working directory. The recommended
command would seem to be `git update-index --refresh`, but in my
testing, this caused "needs update" errors in some of the tests. So I've
resorted to using `git status`, which is probably a slightly more
heavyweight command, but does pass the tests.

If this is a concern, an alternative approach might be to specifically look for the presence of the special all-zero SHA1s returned by `git diff-files`, which indicate that the file is not in sync with the index. Please let me know what you think.

[1]: https://git-scm.com/docs/git-diff-index

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -1065,6 +1065,8 @@ module Git
     # @param [Array] opts the diff options to be used
     # @return [Hash] the diff as Hash
     def diff_as_hash(diff_command, opts=[])
+      # update index before diffing to avoid spurious diffs
+      command('status')
       command_lines(diff_command, opts).inject({}) do |memo, line|
         info, file = line.split("\t")
         mode_src, mode_dest, sha_src, sha_dest, type = info.split
--- a/tests/units/test_status.rb
+++ b/tests/units/test_status.rb
@@ -97,4 +97,20 @@ class TestStatus < Test::Unit::TestCase
       assert(!git.status.untracked?('test_file_2'))
     end
   end
+
+  def test_changed_cache
+    in_temp_dir do |path|
+      git = Git.clone(@wdir, 'test_dot_files_status')
+
+      create_file('test_dot_files_status/test_file_1', 'hello')
+
+      git.add('test_file_1')
+      git.commit('message')
+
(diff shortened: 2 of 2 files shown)


ggilder:
I believe this would also address the issues reported in https://github.com/ruby-git/ruby-git/issues/23

ggilder:
I did a bit more testing after submitting this PR and it looks like `git update-index --refresh` _can_ be used, but since it returns a non-zero exit code if there are files that have truly been modified, we have to ignore the exit code. Let me know if that's more appealing than using `git status`.

ggilder:
😢 
@perlun any chance of a review?

perlun:
Sorry @ggilder, will try to get to this ASAP.

lib/git/lib.rb
```diff
@@ -992,6 +992,8 @@ def command(cmd, opts = [], chdir = true, redirect = '', &block)
     # @param [Array] opts the diff options to be used
     # @return [Hash] the diff as Hash
     def diff_as_hash(diff_command, opts=[])
+      # update index before diffing to avoid spurious diffs
+      command_lines('status')
```
jcouball: It would be more efficient to use `command('status')` here since you are throwing away the results.

jcouball (commented):
@ggilder sorry this has taken so long to merge.  I do have one comment about using `command('status')` instead of `command_lines('status')`.

Since this has been stale so long, if you don't respond in a few days at least to tell me you are still interested in making this change, I'll close this PR and create a new PR that includes your change plus the one I suggested.

Again, sorry for the delay.

ggilder:
@jcouball thanks for taking a look! I made the change you mentioned and rebased on latest master.

jcouball (approved):
Ok, looks good to merge.  Thanks for the contribution!

PR #416, opened 2019-08-02
Allow consumers to point git binary via env var

afiune:

### Description
By adding a new environment variable called `GIT_PATH` we can allow
consumers, that is, a user of a gem that the gem itself uses the git gem,
to customize the location of the git binary.

Example: Having a gem called `git-tool` that uses this gem `git`, if I,
as a user wants to modify the git bin location, I could do:
```
GIT_PATH=/foo/bin git-tool bar
```

Signed-off-by: Salim Afiune <afiune@chef.io>

### DIFF
--- a/lib/git/config.rb
+++ b/lib/git/config.rb
@@ -10,7 +10,7 @@ module Git
     end
 
     def binary_path
-      @binary_path || 'git'
+      @binary_path || ENV['GIT_PATH'] && File.join(ENV['GIT_PATH'], 'git') || 'git'
     end
 
     def git_ssh
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ Git env config
     # If you need to use a custom SSH script
     config.git_ssh = '/path/to/ssh/script'
   end
-
 ```
 
+_NOTE: Another way to specify where is the `git` binary is through the environment variable `GIT_PATH`_
 
 Here are the operations that need read permission only.
 
--- a/tests/units/test_config.rb
+++ b/tests/units/test_config.rb
@@ -31,9 +31,13 @@ class TestConfig < Test::Unit::TestCase
   def test_env_config
     with_custom_env_variables do
       begin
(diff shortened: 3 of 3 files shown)


afiune:
@perlun and @tarcinil -- this PR is ready for review.

lib/git/config.rb
```diff
@@ -10,7 +10,7 @@ def initialize
     end
 
     def binary_path
-      @binary_path || 'git'
+      @binary_path || ENV['GIT_BIN'] || 'git'
```
perlun: How about `GIT_PATH` instead? Or `GIT_BINARY`? (I would prefer the first option, but interested in hearing your opinions also.)

afiune: In my opinion, pointing to the actual binary is the best way to give users the flexibility that they need. Though I can see the confusion since the method name is `binary_path`. I could rename the function if that makes it more clear since it is not returning the path of the binary but the binary itself.

What do you think @perlun?

afiune:
I just renamed the environment variable as you suggested, I can change the behavior to provide the path is this is what you think is best for the users. Open to suggestions.

lib/git/config.rb
perlun: @afiune Thanks. I was just thinking about changing the environment variable name, sorry for not being so clear here. 👍

perlun: Looked at your change now. I think `GIT_PATH` would be the clearer way. Sorry for being slightly picky, but please change the env variable to that and I'll be happy to get this merged ASAP.

afiune: Sounds good. I’ll make the change, update the tests and documentation. 💯

afiune:
@perlun Ready to merge! 💯 

tenor-50737593

afiune:
@perlun ping 😅

perlun:
Looks good @afiune! Sorry for the delays on this. Merging now.

Issue #419, opened 2019-09-17
Add option to disable SSL verification


synthead:
### Subject of the issue
When cloning a git repo over HTTPS, the user currently has no option to ignore SSL verification, if necessary.  The `GIT_SSL_NO_VERIFY` environment variable can be set, but it would be better to also allow the user to explicitly set this in the global settings, like so:

```ruby
  Git.configure do |config|
    # If you need to ignore SSL verification
    config.git_ssl_no_verify = true
  end
```

### Your environment
git 2.23.0
ruby-git 1.5.0
ruby 2.6.4p104

### Steps to reproduce
This is a feature request and cannot be reproduced.

### Expected behaviour
Allow an option to skip SSL verification.

### Actual behaviour
There is no option to skip SSL verification.

PR #420, opened 2019-09-17
Add git_ssl_no_verify option to disable SSL verification for HTTPS repos

synthead:

### Description
Fixes https://github.com/ruby-git/ruby-git/issues/419.

This PR adds a new `git_ssl_no_verify` configuration option like so:

```ruby
  Git.configure do |config|
    config.git_ssl_no_verify = true
  end
```

This option sets `ENV['GIT_SSL_NO_VERIFY']` to `'true'` via `Git::Lib#set_custom_git_env_variables` when `Git::Base.config.git_ssl_no_verify` is truthy.  It allows users to clone git repos over HTTPS without verifying SSL.  This can sometimes ease friction in corporate environments with custom CAs, for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ Git env config
 
     # If you need to use a custom SSH script
     config.git_ssh = '/path/to/ssh/script'
+
+    # If you need to ignore SSL verification
+    config.git_ssl_no_verify = true
   end
 
 ```
--- a/lib/git/config.rb
+++ b/lib/git/config.rb
@@ -2,11 +2,12 @@ module Git
 
   class Config
 
-    attr_writer :binary_path, :git_ssh
+    attr_writer :binary_path, :git_ssh, :git_ssl_no_verify
 
     def initialize
       @binary_path = nil
       @git_ssh = nil
+      @git_ssl_no_verify = nil
     end
 
     def binary_path
@@ -15,6 +16,11 @@ module Git
 
     def git_ssh
       @git_ssh || ENV['GIT_SSH']
+    end
+
+    def git_ssl_no_verify
+      return 'true' if @git_ssl_no_verify
+      ENV['GIT_SSL_NO_VERIFY']
     end
 
   end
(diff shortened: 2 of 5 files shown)


synthead:
Are there any changes you'd like me to make to this PR?

jcouball:
Thank you for this contribution!

One thing I do not like about this is that you set the attribute with a truthy value, but it returns a string.

One might be tempted to set the value to 'false' which would return 'true' when queried later:

```
2.7.2 :001 > c = Git::Config.new
 => #<Git::Config:0x00007f8acf8f7350 @binary_path=nil, @git_ssh=nil, @git_ssl_no_verify=nil>
2.7.2 :002 > c.git_ssl_no_verify
 => nil
2.7.2 :003 > c.git_ssl_no_verify = 'true'
 => true
2.7.2 :004 > c.git_ssl_no_verify
 => "true"
2.7.2 :005 > c.git_ssl_no_verify = 'false'
 => "false"
2.7.2 :006 > c.git_ssl_no_verify
 => "true"
2.7.2 :007 >
```

Also, `c.git_ssl_no_verify` overrides the `GIT_SSL_NO_VERIFY` environment variable ONLY if c.git_ssl_no_verify is set to a non-truthy value.  What is someone wants to use this option to disable the GIT_SSL_NO_VERIFY command?  For this case, I think that you should only fallback to the environment variable if `@git_ssl_no_verify` is unset.

I think two options for this interface that would be better:

1. Allow the user to set the value as a string and don't have an opinion about what is true/false/valid.  It would simply be passed as-is to the command. If unset, use the environment variable.

2. Have an opinion about what is valid and accept only values that git allows to represent a boolean value. Valid values are documented in [the boolean section of git-config](https://git-scm.com/docs/git-config#Documentation/git-config.txt-boolean). If unset, use the environment variable.

jcouball:
Closing this PR without accepting it.

You can set the `http.sslVerify` global config option to accomplish this for all repositories:

```ruby
Git.global_config('http.sslVerify', 'false')
```

For specific repos, you can clone with that option:

```ruby
git.clone(<path>, config: ['http.sslVerify=false'])
```

Alternately, the `GIT_SSL_NO_VERIFY` environment variable still works.

Issue #418, opened 2019-09-11
Diff parsing fails when filenames contain unicode characters


cyclotron3k:
### Subject of the issue
Unicode characters in filenames breaks diffing

### Your environment
* git version 2.20.1
* git (1.5.0)
* ruby 2.6.4p104 (2019-08-28 revision 67798) [x86_64-linux]

### Steps to reproduce
```bash
$ mkdir -p /tmp/test && cd /tmp/test
$ git init
$ printf '\xE2\x98\xA0' # print a skull and crossbones symbol
$ touch 'some_file'
$ git add .
$ git commit -m"init"
$ touch 'my_other_file_\xE2\x98\xA0'
$ git add .
$ git commit -m"commit"
```
```ruby
git = Git.init '/tmp/test'
git.diff('@^').each { |x| ... }
```

### Expected behaviour
The diff to be produced as usual

### Actual behaviour
``NoMethodError: undefined method `[]' for nil:NilClass``

The first line of a diff is usually something like `git --diff a/somefile b/somefile`, but when the filename contains a unicode character, the filenames are escaped, (e.g. `diff --git "a/my_other_file_\\xE2\\x98\\xA0" "b/my_other_file_\\xE2\\x98\\xA0"`) and [this regex](https://github.com/ruby-git/ruby-git/blob/9bd4407c56068e1604f14a1b6c0c5a84868e6378/lib/git/diff.rb#L136) fails to detect the line.

jcouball:
@cyclotron3k if you are still interested, can you see if this issue was fixed with the 1.6.0.pre1 version?  I think this issue might be fixed by #405.

cyclotron3k:
Unfortunately it doesn't seem to have helped:
```
irb(main):005:0> git.diff('@^').each { |x| x }
Traceback (most recent call last):
       10: from /usr/local/bin/irb:23:in `<main>'
        9: from /usr/local/bin/irb:23:in `load'
        8: from /usr/local/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        7: from (irb):5
        6: from (irb):5:in `rescue in irb_binding'
        5: from /usr/local/bundle/gems/git-1.6.0.pre1/lib/git/diff.rb:68:in `each'
        4: from /usr/local/bundle/gems/git-1.6.0.pre1/lib/git/diff.rb:109:in `process_full'
[...]

PR #440, opened 2020-01-31
Fix diffs of files that have quoted paths

jcouball:

### Description
As was reported in #418, if `git diff` reports a file path that has special characters like unicode, backslash, double quotes, etc. then the diff fails.

#### lib/git/diff.rb
The problem was found to be in the diff parser not knowing how to deal with quoted paths.

#### lib/git/lib.rb
Additionally, we always tell git not to quote the paths in most cases so that Unicode characters come through without any special parsing.  This is done by adding the `-c core.quotePath=false` to the global ops of all git commands.  See the details of this option in the [core.quotePath documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath).

#### tests/units/test_diff_with_quoted_path.rb
This test runs a diff containing two files: one whose path contains a backslash and double quote and one whose path with a non-ascii, unicode character in the filename.  The diff should list first path with quotes and the second path without quotes like this:

```
$ git -c core.quotePath=false diff -p '@^'
diff --git "a/asdf\"asdf" "b/asdf\"asdf"
new file mode 100644
index 0000000..e69de29
diff --git a/my_other_file_☠ b/my_other_file_☠
new file mode 100644
index 0000000..e69de29
$
```

Note the different between the lines beginning with `diff --git`.

#### tests/files/quoted_path/**/*

This is the test repository who last commit has two changes: one to `asdf\"asdf` and one to `my_other_file_☠`.

#### tests/units/test_logger.rb

Since the `-c core.quotePath` was added to every command line, the logger tests had to be changed to ignore changes in global options.


jcouball:
Using PR #449 for the changes originally in this PR.

PR #448, opened 2020-02-09
Add YARD documentation to ruby-git

jcouball:
### Your checklist for this pull request
🚨Please review the [guidelines for contributing](https://github.com/ruby-git/ruby-git/blob/master/CONTRIBUTING.md) to this repository.

- [ ] Ensure all commits include DCO sign-off.
- [ ] Ensure that your contributions pass unit testing.
- [ ] Ensure that your contributions contain documentation if applicable.

### Description

Create YARD documentation for ruby-git instead of rdoc. Make improvements to the documentation.

PR #451, opened 2020-02-12
Update commit to return an object, matching the documentation

jamesiarmes:
Signed-off-by: James Armes <jamesiarmes@gmail.com>

### Description
According to the [documentation](https://www.rubydoc.info/gems/git/1.6.0/Git/Base#commit-instance_method), both the `commit` and `commit_all` methods should return an object, but instead they return the output from running `git commit` as a string. This pull requests update these methods to return an object, matching the documentation.

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -291,6 +291,8 @@ module Git
     #
     def commit(message, opts = {})
       self.lib.commit(message, opts)
+
+      gcommit('HEAD')
     end
         
     # commits all pending changes in the index file to the git repository,
@@ -299,6 +301,8 @@ module Git
     def commit_all(message, opts = {})
       opts = {:add_all => true}.merge(opts)
       self.lib.commit(message, opts)
+
+      gcommit('HEAD')
     end
 
     # checks out a branch as the new git working directory
--- a/tests/units/test_base.rb
+++ b/tests/units/test_base.rb
@@ -91,9 +91,11 @@ class TestBase < Test::Unit::TestCase
 
       base_commit_id = git.log[0].objectish
 
-      git.commit("Test Commit")
+      o = git.commit("Test Commit")
+      assert(o.is_a?(Git::Object::Commit))
 
       original_commit_id = git.log[0].objectish
+      assert(o.sha == original_commit_id)
 
       create_file('test_commit/test_file_3', 'content test_file_3')
       


jcouball:
hmm, commit and commit_all **do** return an Object.  It just so happens that the object is a String.

The documentation could definitely use some improvement.  I have gotten that started in #448.

I am hesitant to change an interface that has been in place for so long with out a major version bump.

Thoughts?

jamesiarmes:
To be clear, the documentation indicates it should return a `Git::Object` (which `Git::Object::Commit` is), rather than any ruby object. I would argue that anybody expecting a string at this point would be developing against an undocumented "feature" and shouldn't expect that to remain the case.

Having said that, I understand the concern about breaking existing APIs without a major version bump, so I can see arguments for both.

While this change does more accurately reflect the documentation and I feel is more intuitive than the current return value, I would be fine holding it for the next major version (I would still like to see this in a future version) with the caveat that the documentation be updated to reflect the current return value (which I would be happy to contribute to).

jcouball:
I wonder if the current state of generated docs for this project (where by default, everything returns a `Git::Object`) represents a bug in Yard?  [Yard documents](https://rubydoc.info/gems/yard/file/docs/WhatsNew.md#Added_yardoc_--default-return_and_yardoc_--hide-void-return__0_5_0_) that:

> YARD defaults to displaying (Object) as the default return type of any method that has not declared a @return tag

However, it appears this `Object` is evaluated within the main module of the project (for example, the `Git` module for this project).  Because of this, the return type `Object` is linked to `Git::Object` not because the documentation changed, but merely by the presence of `Git::Object`. 

Nesting the `Object` class into a submodule does not have this behavior.

Removing `Git::Object` results in YARD documenting `Object` without a link, implying `::Object`.

jcouball:
What do you think about this overall plan to deal with this issue?  These steps would be done in this order:

1. Immediately add a `.yardopts` file that sets the default return type of any method that has not declared a `@return` tag to `::Object`.  This can be done by including `--default-return ::Object` in the .yardopts file.  That is my best guess as to the original intent.
2. Over time, explicitly document the entire API for this gem using YARD
3. Eventually reconsider what is returned for each method like `Git::Base#commit`

jcouball:
Per the last comment, `.yardopts` has been configured not to document a return value when a `@return` directive has not been given for a method.  Instead of `::Object`, I have configured `.yardopts` to consider undocumented return values to be `void` which means that YARD should not show the return type in the documentation.

PR #461, opened 2020-03-27
Git::Lib#normalize_encoding early return fix

sd-trailhead-james:

### Description
`Git::Lib#normalize_encoding`: when checking to see if the given string
is already in the desired encoding, fix the comparison check to compare
string value to string value, not class to string value.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -1050,7 +1050,7 @@ module Git
     end
 
     def normalize_encoding(str)
-      return str if str.valid_encoding? && str.encoding == default_encoding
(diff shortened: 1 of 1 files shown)


sd-trailhead-james:
I'm seeing lots of Ruby 2.7 deprecation warnings in the two `String#encode` calls that take place after the single line that this PR fixes. Upon closer inspection I realized that those encode lines should never be getting called, because the string is already in my desired encoding. This PR fixes a bug in the early return check to get the early return to work properly if the string is already in the desired encoding.

PR #463, opened 2020-04-04
Add support for git 2.25 and ruby 2.7

terceiro:
I made these changes while preparing a new upload of ruby-git for Debian.

The last commit is not necessarily related to supporting the new git and ruby, but it has no effect for upstream developers while making the life of downstream maintainers easier. Please consider accepting it as well, but if not I can drop it from this PR.


jcouball:
The Ruby 2.7 warning was already submitted in #453.  I have merge that PR to master.

Can you rebase and then remove that from this PR?

PR #470, opened 2020-06-05
Support multiple --config options in Git.clone

svl7:
Makes opts[:config] in Git.clone accept an Array of <key>=<value> pairs
so that multiple --config options can be passed to the clone command.
A single config entry can still be passed as a String.

Signed-off-by: Marco Rüdisüli <git@rudisu.li>

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,9 @@ And here are the operations that will need to write to your git repository.
         { :repository => '/opt/git/proj.git',
            :index => '/tmp/index'} )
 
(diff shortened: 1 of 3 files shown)


svl7:
Just here because I got the stale bot message, maybe a maintainer can merge or decline this? @jcouball

bond-os:
It would be nice with this PR being merged to next release.

jcouball:
Since I need to make a change or two to the implementation and documentation, I have re-implemented this PR in #653

PR #472, opened 2020-06-22
Support core.abbrev 7-40

SocalNick:

### Description
We ran into an issue when we bumped to Ubuntu 18 (bionic) and started
seeing failures in process_full_diff because the index lines had 11
character SHA abbreviations. This should allow core.abbrev to be set to
anything between 7 (the default) and 40 (the max).


lib/git/diff.rb
```diff
@@ -89,10 +89,10 @@ def binary?
       end
 
       def blob(type = :dst)
-        if type == :src
-          @base.object(@src) if @src != '0000000'
-        else
-          @base.object(@dst) if @dst != '0000000'
+        if type == :src && !/0{7,40}/.match(@src)
```
SocalNick: Perhaps worth mentioning, I couldn't find a test that required the check for `@src != 0000000` to be there. Assuming there's some edge case it's needed for, I changed it to the regex version.

lib/git/diff.rb
```diff
@@ -89,10 +89,10 @@ def binary?
...
+          @base.object(@src)
+        elsif !/0{7,40}/.match(@dst)
```
ngan: Might be good to pull the regex out to a constant and use `match?` (instead of `match`) since you don't care to capture any values.

SocalNick: Updated to use a constant for the regex. I initially tried `match?` but that isn't available in Ruby 2.3 which is in the test matrix for this project.

ngan:
@jcouball any chance we can get this landed and released?

tomgilligan:
i should have checked here first, ended up with my own version of a fix https://github.com/tomgilligan/ruby-git/commit/2d82185b482302efed0e19fd1b1aeee73d318715

worth mentioning that the minimum abbrev is apparently 4
https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev

tomgilligan:
also i think `NIL_BLOB_REGEXP` should be 'anchored'

lib/git/diff.rb
```diff
@@ -72,6 +72,7 @@ def each(&block) # :yields: each Git::DiffFile in turn
     class DiffFile
       attr_accessor :patch, :path, :mode, :src, :dst, :type
       @base = nil
+      NIL_BLOB_REGEXP = /0{7,40}/.freeze
```
ngan:
```suggestion
      NIL_BLOB_REGEXP = /\A0{7,40}\z/.freeze
```

lib/git/diff.rb
tommy-gilligan: seconded, there is no test for `@src` check

SocalNick:
@tomgilligan thanks for the feedback! Ngan committed the change to anchor the `NIL_BLOB_REGEXP` and I added test to ensure this works with 4,7, and 40 (really anything between 4-40) for `core.abbrev`.

Not sure why the CI check isn't running, otherwise, is this good to merge?

jcouball:

I have this one in my queue for review.  Sorry it has taken so long.

ngan:
@jcouball bump

jcouball (approved):
Looks good.  Thank you for the thorough update!

PR #473, opened 2020-06-24
Allows a logger to be given during cloning

sergio-bobillier:

🚨 Please review the [guidelines for contributing](https://github.com/ruby-git/ruby-git/blob/master/CONTRIBUTING.md) to this repository.

- [x] Ensure that your contributions contain documentation if applicable. (N/A)

### Description

Currently, if you pass a logger to the `clone` method in the Base class it is completely ignored, not only it is not used during the cloning phase but the resulting `Base` instance doesn't have the logger attached to it, meaning that all subsequent operations over that repository are not logged.

This Pull Request attempts to remedy that by adding code to pass the logger object to the constructor of the `Git::Lib` class as well as to the constructor of the `Base` class itself.

In order to avoid coupling the classes (more than they already are) I decided to merge the logger to the Hash returned by the `clone` method of the `Git::Lib` class instead of passing it and then having the method return it in the resulting `Hash`.

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -25,10 +25,11 @@ module Git
     #    :index_file
     #
     def self.clone(repository, name, opts = {})
-      # run git-clone 
-      self.new(Git::Lib.new.clone(repository, name, opts))
+      # run git-clone
+      log = opts[:log]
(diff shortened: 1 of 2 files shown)


jcouball:
This was fixed in #501

PR #482, opened 2020-08-11
Add commit --allow-empty-message option and fix empty message parsing in process_commit_log_data

### LINKED ISSUE #481: Incorrect log if repo contains commits with empty message
### Subject of the issue
If repository contains commit made by `git commit --allow-empty --allow-empty-message -m ""` log return incorrect commits history. Returned history starts from this commit with empty message.

### Your environment
```
$ git --version
git version 2.20.1
```
```
$ sudo gem install git     
Successfully installed git-1.7.0
```
```
$ ruby --version 
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-linux-gnu]
```

### Steps to reproduce
Create testing git repository:
```
mkdir test
cd test
git init
git commit -m "test" --allow-empty
git commit -m "test2" --allow-empty
git commit -m "" --allow-empty --allow-empty-message
git commit -m "test3" --allow-empty
```
Get real commit history:
```
$ git log | tee test
commit fc940fc8858011566d6a16076d01bcf189955b37
Author: ...
Date:   Mon Aug 10 17:34:10 2020 +0300

    test3

commit 58dc8ed01e328ea4cea39b28a01aa4500be5e385
Author: ...
Date:   Mon Aug 10 17:34:02 2020 +0300

commit 458f6e6b9305df15a9c23c3dffaf7448d917c72f
Author: ...
Date:   Mon Aug 10 17:33:43 2020 +0300

    test2

commit 9dd305fe98ef19385906e39cdb31ee7696f53e03
Author: ...
Date:   Mon Aug 10 17:33:37 2020 +0300

    test
```
Create test ruby script (named test.rb):
```ruby
require "git"

g = Git.open("~/test")

g.log.to_a.each do |com|
   puts com.sha 
end
```
Call it:
```
$ ruby test.rb                                                 
fc940fc8858011566d6a16076d01bcf189955b37
58dc8ed01e328ea4cea39b28a01aa4500be5e385
```
As we can see this is not all commit history.

### Expected behaviour
Retrieving complete commit history

### Actual behaviour
The commit history is truncated from a commit with an empty message

kpd200081:

### Description
If the commit doesn't contain a message, parsing the output of "git log"
results in an error, which occures because the end of the commit message
can't be detected.

fixes #481

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -243,6 +243,8 @@ module Git
           next
         end
 
+        in_message = false if in_message && line[0..3] != "    "
+
         if in_message
           hsh['message'] << "#{line[4..-1]}\n"
           next
@@ -559,9 +561,10 @@ module Git
     #  :author
     #  :date
     #  :no_verify
+    #  :allow_empty_message
     #
     # @param [String] message the commit message to be used
-    # @param [Array] opts the commit options to be used
+    # @param [Hash] opts the commit options to be used
     def commit(message, opts = {})
       arr_opts = []
       arr_opts << "--message=#{message}" if message
@@ -571,6 +574,7 @@ module Git
       arr_opts << "--author=#{opts[:author]}" if opts[:author]
       arr_opts << "--date=#{opts[:date]}" if opts[:date].is_a? String
       arr_opts << '--no-verify' if opts[:no_verify]
+      arr_opts << '--allow-empty-message' if opts[:allow_empty_message]
 
       command('commit', arr_opts)
     end
--- a/tests/units/test_commit_with_empty_message.rb
+++ b/tests/units/test_commit_with_empty_message.rb
@@ -1,0 +1,25 @@
+#!/usr/bin/env ruby
+require File.dirname(__FILE__) + '/../test_helper'
+
+class TestCommitWithEmptyMessage < Test::Unit::TestCase
+  def setup
+    set_file_paths
+  end
+
+  def test_without_allow_empty_message_option
+    Dir.mktmpdir do |dir|
+      git = Git.init(dir)
+      assert_raises Git::GitExecuteError do
+        git.commit('', { allow_empty: true })
+      end
+    end
+  end
+
+  def test_with_allow_empty_message_option
+    Dir.mktmpdir do |dir|
+      git = Git.init(dir)
+      git.commit('', { allow_empty: true, allow_empty_message: true})
+      assert_equal(1, git.log.to_a.size)
+    end
+  end
+end
--- a/tests/units/test_log.rb
+++ b/tests/units/test_log.rb
@@ -78,5 +78,17 @@ class TestLog < Test::Unit::TestCase
       @git.log.object('no-exist.txt').size
     end
   end
-  
+
+  def test_log_with_empty_commit_message
+    Dir.mktmpdir do |dir|
+      git = Git.init(dir)
+      expected_message = 'message'
+      git.commit(expected_message, { allow_empty: true })
+      git.commit('', { allow_empty: true, allow_empty_message: true })
+      log = git.log
+      assert_equal(2, log.to_a.size)
+      assert_equal('', log[0].message)
+      assert_equal(expected_message, log[1].message)
+    end
+  end
 end


kpd200081:
@rvodden @perlun @tarcinil, please can somebody review this pull request?

PS: build passed https://travis-ci.org/github/ruby-git/ruby-git/builds/716900598, I'm not sure why the status is not updating

jcouball:

jcouball:
@kpd200081 can you please add a failing test that this change fixes?

You should probable create your own test fixture instead of changing existing fixtures. If you don't have time to figure this out, I can help out -- just let me know.  Here is what I told someone else:

> Reusing the default test fixture (test/files/working/dot_git) breaks tests that don't have anything to do with the change you are making. To me this is tech debt that is piling up with each change like this one. The next PR might have to correct all the same tests plus the new ones introduced in this PR. Additionally, I think merge conflicts will be more of a problem if everyone is trying to change the same test fixture for their tests.
> 
> You can see an example of using a different fixture in `tests/units/test_diff_non_default_encoding.rb` for using the fixture `tests/files/encoding`.  It would probably be good to create a new git working copy setup just for your test instead of copying an existing one that has a lot of setup for other things.

kpd200081:
@jcouball, I'm not sure if a test is needed for this case. 

Anyway, I'm not familiar with Ruby tests, so if you think it's necessary it would be nice if you could help with that.

kpd200081:
@jcouball I added a test that falling without this fix, but I had to change the `commit` method a bit, take a look please

As we can see in travis it fail because run on published version of gem (I don't know why, locally I have same behavior)

lib/git/lib.rb
```diff
@@ -571,6 +574,7 @@ def commit(message, opts = {})
       arr_opts << "--author=#{opts[:author]}" if opts[:author]
       arr_opts << "--date=#{opts[:date]}" if opts[:date].is_a? String
       arr_opts << '--no-verify' if opts[:no_verify]
+      arr_opts |= opts[:raw_opts] if opts.key?(:raw_opts)
```
jcouball: It seems like the convention (and I'm not giving an opinion if the convention is good or bad) is to add an option `:allow_empty_message` instead of passing in an array of "raw options" to allow for this option.

What do you think about sticking with that convention and changing this line to (this is my untested guess):

```Ruby
arr_opts << '--allow_empty_message' if opts[:allow_empty_message]
```

and then update the documentation accordingly.

Also, isn't the documentation for `opts` wrong?  Should it say that a Hash is expected instead of an Array?

tests/units/test_log_empty_message.rb
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+require File.dirname(__FILE__) + '/../test_helper'
+
+class TestLogEmptyMessage < Test::Unit::TestCase
+  def setup
+    set_file_paths
+  end
+  
+  def test_repo_with_empty_commit
```
jcouball: I think this test name should be changed to `test_repo_with_empty_commit_message` since empty commit is something different.

For this functionality, I would think to do two tests:
1. 
a test that does not have a commit message and does not use the `:allow_empty_message` option
2. a test that does not have a commit message and DOES use the `:allow_empty_messages` option

Since the `:allow_empty` and `:allow_empty_message` options are so similar, I found the use of the `:allow_empty` option in this particular test confusing. I'll leave it up to you if you want to do anything about this.

kpd200081: >I think this test name should be changed to `test_repo_with_empty_commit_message` since empty commit is something different.

oh, year, I did misspell here

>Since the :allow_empty and :allow_empty_message options are so similar, I found the use of the :allow_empty option in this particular test confusing. I'll leave it up to you if you want to do anything about this.

`:allow_empty` just allow commit without changes, so it was added to both commits

lib/git/lib.rb
kpd200081: I think about it and decide that `:allow_empty_message` will be really rare used, at the other hand `:raw_opts` can be used for different purposes

>Also, isn't the documentation for opts wrong? Should it say that a Hash is expected instead of an Array?

Year, it definitely Hash, should we fix it in this Pull Request?

>It seems like the convention

It's not convention, just array appending

tests/units/test_log_empty_message.rb
kpd200081: >a test that does not have a commit message and does not use the `:allow_empty_message` option

we can't commit without this option, this results in an error in `git` and test failure

lib/git/lib.rb
jcouball: I believe we should change this to follow the convention of using specific hash keys for passing options (passing `:allow_empty_message` instead of `:raw_opts`)

One reason for this convention is to constrain how use this gem is used so users don't give combinations of options that this gem wouldn't respond well to. This is how the rest of the Git gem interface works. I don't see a reason to violate this convention just for this case. After all, isn't `:allow_empty` just as rarely used?  Why give two different ways for passing arguments?

kpd200081:
@jcouball I apply suggestions to test name and `opts` type documentation, also I reply to other questions.

lib/git/lib.rb
kpd200081: Ok, you convinced me

tests/units/test_log_empty_message.rb
kpd200081: So, as test without `:allow_empty_message` can't work because of `git` internal error, can we resolve this?

kpd200081:
@jcouball should I rebase it?

kpd200081:
@jcouball I fix commit author and rebase to remove merge commit, now  I think nothing block from accepting PR.

PR #485, opened 2020-09-02
Allow push options

liger1978:
Signed-off-by: harbottle <harbottle@room3d3.com>

- [ ] Ensure all commits include DCO sign-off.
- [ ] Ensure that your contributions pass unit testing.

### Description
Accept array of push_options in opts to "push" method: https://git-scm.com/docs/git-push#Documentation/git-push.txt---push-optionltoptiongt


lib/git/lib.rb
```diff
@@ -789,6 +789,9 @@ def push(remote, branch = 'master', opts = {})
       arr_opts << '--mirror'  if opts[:mirror]
       arr_opts << '--delete'  if opts[:delete]
       arr_opts << '--force'  if opts[:force] || opts[:f]
+      if opts[:push_options]
+       arr_opts.concat(opts[:push_options].map { |push_option| ['--push-option', push_option] }.flatten)
+      end
```
jcouball, line 831: When I look at the rest of `Git::Lib`, I notice the mapping from `opts` to `arr_opts` where the `opts` value can multi-valued is inconsistently implemented. 

What do you think about providing a common way of doing multi-valued options:

```Ruby
def multi_valued_opt(opt, values)
  Array(values).map { |value| [opt, value] }.flatten
end
```

This implementation uses `Array(...)` so that the value for (for example) `opts[:push_options]` can be either a single object or an Array or any object that knows how to convert itself into an array.

and then implementing `--push-option` with this line:

```Ruby
arr_opts += multi_valued_opt('--push-option', opts[:push_options]) if opts[:push_options]
```

The hope would be that other instances of this pattern would adopt `multi_valued_opt` to make their code more understandable.

liger1978:
@jcouball sounds sensible. Can you merge this in the meantime though please?

jcouball:
Fair enough.  It might be better to introduce `multi_valued_opt` in a separate PR. In order to accept this PR, you will need to (1) add a test, (2) add documentation, and (3) squish commits into one commit.

For testing, I'd expect just one test that makes sure the command line gets built correctly for `push` when the `:push_options` option is given. This will help when the next PR transforms the implementation to use something like `multi_valued_opt`.

For documentation, updating the README.md with an example is sufficient.  Adding full YARD documentation to Git::Lib#push would be a bonus.

liger1978:
I don't have time to do all the tests and docs you want, so closing this. I have forked for my purposes.

PR #487, opened 2020-09-07
Simplify git commands

hatkyinc2:
### Description
`git -C` to replace the confusing pair of `--git-dir` and `--work-tree`

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -993,8 +993,7 @@ module Git
 
     def command(cmd, opts = [], chdir = true, redirect = '', &block)
       global_opts = []
-      global_opts << "--git-dir=#{@git_dir}" if !@git_dir.nil?
-      global_opts << "--work-tree=#{@git_work_dir}" if !@git_work_dir.nil?
+      global_opts << ["-C", "#{@git_work_dir}"] if !@git_work_dir.nil?
       global_opts << ["-c", "color.ui=false"]
 
       opts = [opts].flatten.map {|s| escape(s) }.join(' ')
--- a/tests/units/test_logger.rb
+++ b/tests/units/test_logger.rb
@@ -19,7 +19,7 @@ class TestLogger < Test::Unit::TestCase
     @git.branches.size
     
     logc = File.read(log.path)
-    assert(/INFO -- : git '--git-dir=[^']+' '--work-tree=[^']+' '-c' 'color.ui=false' branch '-a'/.match(logc))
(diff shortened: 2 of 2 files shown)


jcouball:
I think having both `--git-dir` and `--work-tree` options is necessary to cover the case where the `.git` directory isn't located within the work tree or isn't named `.git`.

It seems like we should have tests for that but I guess we don't since the tests passed with this change.

hatkyinc2:
the `.git` being a file in place of a directory is exactly the reason I started down the rabbit hole, I believe it currently has bugs, for both submodules and worktrees, opening a Git and trying to checkout a new branch fails. Even if the submodules code is to be fixed it should be duplicated or moved to the initialize function.
I'm in too deep and I'm a bit lost for a few days already...

The two switches need to be amended differently for each case and I'm yet to fully get exactly how they work, couldn't find much documented to explain what they mean.
I also get into weird states where the change tree doesn't match the preped files or a .lock file missing issue...
So the `-C` is close to an actual change directory, which actually works in place of the two switches for all cases. `git` CLI just figures them alone and the rest of the stuff.. you may still override them even after a `-C` when they would be relative to that (assuming relative paths are used over full ones)

hatkyinc2:
I'm happy to add tests for these, due they currently fail for other stuff...
If you want to depend this change merge on added tests for submodule and worktree, that's fine.

hatkyinc2:
So this is a submodule test it fails for both master and here 😭 
https://github.com/ruby-git/ruby-git/pull/488

But my conclusions from it are:
* This code is probably not useful https://github.com/ruby-git/ruby-git/blob/master/lib/git/base.rb#L58-L65
* The separation of `--git-dir` and `--work-tree` over `-C` doesn't help currently

hatkyinc2:
This is a test for worktree new branch checkout, same error as with the submodule
https://github.com/ruby-git/ruby-git/pull/489

hatkyinc2:
I think at this point I need help.
Like I don't get it, if I do exactly as the test and the command listed on the terminal it works
But obviously this doesn't work.

It's probably something super silly, that's how it usually works :(

jcouball:
@hatkyinc2 I was wondering if you could take a look at #490.  In addition to fixing a bug (how the default for `index` is calculated) it also adds a test for when `git-dir` is in a non-standard place relative to `work-tree`.  This test should fail with the change you are proposing in this PR.

jcouball:
Closing this for the reasons stated in the comments above.  It comes down to the matter that sometimes the `.git` file/directory is not within the working copy.

PR #502, opened 2020-12-23
Add YARD documentation to ruby-git

jcouball:

### Description
Create YARD documentation for ruby-git instead of rdoc. Make improvements to the documentation.

Not everything is documented yet. Additional PRs will add YARD doc for everything.

The work done so far shows what this would look like for most methods in the Git module like Git.bare, Git.open, Git.clone, Git.init.

In order to see this updated documentation, you will need to clone this PR branch and then run:

```
bundle install
bundle exec rake yard && open doc/index.html
```

### JRuby Not Supported for Generating YARD docs

This project uses the `redcarpet` gem to be able to use Markdown to format YARD documentation.  `redcarpet` is not compatible with JRuby.

To make this clear, the rake tasks for YARD documentation are not added under JRuby and the `redcarpet` gem is not development dependency when running in JRuby.

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -1,31 +1,23 @@
 require 'git/base/factory'
 
 module Git
-
+  # Git::Base is the main public interface for interacting with Git commands.
+  #
+  # Instead of creating a Git::Base directly, obtain a Git::Base instance by
+  # calling one of the follow {Git} class methods: {Git.open}, {Git.init},
+  # {Git.clone}, or {Git.bare}.
+  #
   class Base
-
     include Git::Base::Factory
 
-    # opens a bare Git Repository - no working directory options
-    def self.bare(git_dir, opts = {})
-      self.new({:repository => git_dir}.merge(opts))
+    # (see Git.bare)
+    def self.bare(git_dir, options = {})
+      self.new({:repository => git_dir}.merge(options))
     end
 
-    # clones a git repository locally
-    #
-    #  repository - http://repo.or.cz/w/sinatra.git
-    #  name - sinatra
-    #
-    # options:
-    #   :repository
-    #
-    #    :bare
-    #   or
-    #    :working_directory
-    #    :index_file
-    #
-    def self.clone(repository, name, opts = {})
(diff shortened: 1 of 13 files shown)


lib/git/base.rb
```diff
@@ -35,56 +27,82 @@ def self.config
...
-    # options:
-    #  :bare
-    #  :index
-    #  :repository
-    #
-    def self.init(working_dir, opts = {})
-      opts[:working_directory] ||= working_dir
-      opts[:repository] ||= File.join(opts[:working_directory], '.git')
+    # (see Git.init)
+    def self.init(directory, options = {})
```
jcouball, line 31: The arg `opts` was changed to `options` to be the same as `Git.init` so the `(see Git.init)` YARD directive would work.

lib/git/base.rb
```diff
@@ -35,56 +27,82 @@ def self.config
...
+      self.new(options)
     end
 
-    # opens a new Git Project from a working directory
-    # you can specify non-standard git_dir and index file in the options
-    def self.open(working_dir, opts={})
-      opts[:working_directory] ||= working_dir
-      opts[:repository] ||= File.join(opts[:working_directory], '.git')
+    # (see Git.open)
+    def self.open(working_dir, options={})
```
jcouball, line 66: The arg opts was changed to options to be the same as `Git.open` so the `(see Git.open)` YARD directive would work.

lib/git/base.rb
```diff
@@ -198,46 +216,53 @@ def lib
...
-      else
-        self.lib.add('.', args[0]||{})
-      end
+    # @option options [boolean] :all
+    #   Update the index not only where the working tree has a file matching
+    #   <pathspec> but also where the index already has an entry.
+    #   See [the --all option to git-add](https://git-scm.com/docs/git-add#Documentation/git-add.txt--A)
+    #   for more details.
+    #
+    def add(paths = '.', **options)
```
jcouball, line 264: Change the signature from `*args` so the individual options can be documented.  I am confident that this is equivalent because the unit tests still pass.

PR #517, opened 2021-03-23
Fix use of spaces instead of tab in lib/git/base.rb

yancyribbens:
This PR fixes a minor oversight in the text spacing.


jcouball (approved):
Thanks @yancyribbens! I appreciate the correction.

PR #519, opened 2021-04-22
Add "g.current_branch" to Readme.md

HatsuMora:
### Description
In my opinion .current_branch is a fairly common operation, and it would be nice to have it listed on the readme file.


jcouball (approved):
Totally agree!  Thank you.

PR #529, opened 2021-07-12
Add -ff option to git clean

cwalsh:
From `git clean --help`:
> Git will refuse to modify untracked nested git repositories (directories with a .git subdirectory) unless a second -f is given.

This adds the ability to pass `-ff` to `git clean` to enable removing directories with a .git subdirectory.

An example use case is when a gem from a git source is cached in vendor/cache/some_gem. Removing would require `git clean -dff`


jcouball (approved):
Looks great! Thank you for the contribution.

PR #555, opened 2022-01-05
Delete tempfile so files at this path are not deleted during tests

bjfish:

### Description
This allows the truffleruby-head CI to pass.  I believe this is currently failing because the `tempfile_object` is being finalized and deleting the file at the corresponding path while the file at the path is still being used during the test.  The problem here is similar to what is described here: https://github.com/ruby-git/ruby-git/blob/735b0833f54fd74a1584070ff22611a6e1a7611f/lib/git/lib.rb#L984.

This solution holds the tempfiles in an array to ensure they are not removed until the `Array#clear` method is called after the test is run.

Alternative solutions:
- call `tempfile_object.close!` - This would prevent the issue but the path wouldn't be cleaned up after the test.  The paths could be tracked and removed manually.


jcouball (approved):
Nice! I noticed the occasional failures. I am glad you stepped up with a solution. What you did makes a lot of sense. Thank you.

Issue #551, opened 2022-01-02
Determine policy for which versions of git are supported


jcouball:
Currently, this project [claims to support git 1.6.0.0 and beyond](https://github.com/ruby-git/ruby-git/blob/master/git.gemspec#L27). However, this claim is not testable since I have been unable to figure out a way to add versions of git that old to the CI build.

So we have to determine:
* What version of git SHOULD we claim to support?
* Going forward, what is the policy for how the supported version of git changes?
* If we say we support a version of git, what responsibility that that imply? e.g. what versions of git should be used in the CI build?

PR #563, opened 2022-01-17
Add support for fetch options "--force/-f" and "--prune-tags/-P".

### LINKED ISSUE #562: Please add support fetch options --prune-tags/-P and --force/-f
### Subject of the issue
fetch is lacking support for `--prune-tags` and `--force`. I'm currently getting an error when I fetch from a remote where a tag has been moved to a new revision. To sort this out, one has to `git fetch --tags --force`

### Steps to reproduce
Try to call fetch with said options

### Expected behaviour
They should work

### Actual behaviour
They don't work (since there's no code to pass them to the git executable)

dhs-rec:
Signed-off-by: Dirk Heinrichs <dheinric@opentext.com>

### Description
Fixes #562.


jcouball (approved):
Tests are missing... I'll add tests for fetch options later.

Also, this project claims to support git 1.6.0 and later which is probably super unrealistic. Feel free to weigh in on #551 if you have an idea for how we should choose which version of Git we support.

PR #566, opened 2022-03-15
Fix bug when grepping lines that contain numbers surrounded by colons

4-Eyes:
Signed-off-by: James Fairbairn <jamesaf112@gmail.com>

### Description

Recently I discovered a bug where the `grep` function was not parsing certain lines correctly. In particular, lines that contained a digit surrounded by colons. For example, take the following file:

```txt
example :20:
```

If you called `grep('example')`, it would return incorrect file paths including the `example ` and the line number would be `20`.

Upon exploring this issue, I discovered that it was due to the regex used for parsing. Effectively, the first capture group needed to be greedy to fix the issue.


jcouball:
@4-Eyes I assume you meant that "the first capture group needed to be non-greedy to fix the issue."

jcouball (approved):
Thanks for the contribution!

At some point we need to move away from fixtures like `test/files/working` because they cause so many files to have to be changed when a test case is added.

4-Eyes:
> @4-Eyes I assume you meant that "the first capture group needed to be non-greedy to fix the issue."

Yes, I did mean that. I'm glad you understood me regardless.

PR #580, opened 2022-05-14
Update README.md

mslinn:
### Description
Corrected docs. No array is returned, instead, an enumerator is returned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ g.index.writable?
 g.repo
 g.dir
 
-g.log   # returns array of Git::Commit objects
+g.log   # returns a Git::Log object, which is an Enumerator of Git::Commit objects
 g.log.since('2 weeks ago')
 g.log.between('v2.5', 'v2.6')
 g.log.each {|l| puts l.sha }


README.md
```diff
@@ -108,7 +108,7 @@ g.index.writable?
 g.repo
 g.dir
 
-g.log   # returns array of Git::Commit objects
+g.log   # returns Enumberator of Git::Commit objects
```
jcouball: "Enumberator" is misspelled.

Maybe it would be better to say: "returns a Git::Log object which is an Enumerator of Git::Commit objects"

jcouball (changes requested):
Thank you for this correction! I have requested a change in your comment.

README.md
mslinn: I thought I had fixed the spelling error :(

Your suggestion is an improvement and should be used.

jcouball: If you agree with my suggestion, please redo this commit and I'll make sure it is reviewed and merged.

mslinn: Just did that

jcouball:
I appreciate the update! It looks great! Can you squash these commits into one commit?

Here is how I would do it on your branch (should only take a few moments):

This assumes that you don't have any other changes and you are on your PR branch:

```shell
# Undo your two commits saving the changes in your work area
git reset HEAD~2
# Add the changes back, commit, and force push the result (with only one commit)
git add .
git commit -m 'Update README.md'
git push --force
```

README.md
mslinn: Squished

jcouball (approved):
Looks great!  Thanks for noticing the error and fixing it!

Issue #574, opened 2022-04-19
Diff parsing fails with opaque error when SSH warnings preceed the standard header


aalong-tr:
### Subject of the issue
I'm experiencing a transient error in a CI environment using the Danger gem. The exception and relevant stack trace:
```
undefined method `[]' for nil:NilClass
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/git-1.10.2/lib/git/diff.rb:148:in `block in process_full_diff'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/git-1.10.2/lib/git/diff.rb:131:in `each'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/git-1.10.2/lib/git/diff.rb:131:in `process_full_diff'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/git-1.10.2/lib/git/diff.rb:110:in `process_full'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/git-1.10.2/lib/git/diff.rb:68:in `each
```

I have identified the root cause as `git diff` output that includes an SSH key warning:
```
Warning: Permanently added the ECDSA host key for IP address '000.00.000.0' to the list of known hosts.
diff --git a/some/file/name b/some/file/name
```
(IP address obscured) 

While SSH warnings are not necessarily the responsibility of this gem, better error handling with explicit explanation of why the `git diff` output did not match expected formatting would make this style of error significantly easier to identify and fix. 

### Your environment
* ruby-git version 1.10.2
* Ruby version 2.7.0

aalong-tr:
Proposed fix in https://github.com/ruby-git/ruby-git/pull/585

jcouball:
Fixed with #684 and released in [git-2.0.0-pre1](https://rubygems.org/gems/git/versions/2.0.0.pre1)

aalong-tr:
Fabulous news, thank you so much for this refactor!

PR #585, opened 2022-06-14
Proof of concept fix for git diff parsing failure on warn [Issue #574]

aalong-tr:

### Description
SSH warnings and other unexpected output from `git diff` can create cryptic errors, as the diff parser assumes that the first line of a diff will always provide a file name for consideration. To mitigate this, ignore lines with unexpected formatting until the first line with a file header is reached. 

Given that the problem code is deep in the internals of diff.rb, I have taken a somewhat aggressive white-box testing approach to my accompanying unit test, which I acknowledge may prove brittle if the internals of the class change. I am open to alternative strategies as per the maintainers' recommendations. 

Prospective fix for Issue #574

### DIFF
--- a/lib/git/diff.rb
+++ b/lib/git/diff.rb
@@ -133,6 +133,7 @@ module Git
             current_file = Git::EscapedPath.new(m[2]).unescape
             final[current_file] = defaults.merge({:patch => line, :path => current_file})
           else
+            next unless current_file # Protects against warnings or other garbage output preceeding the diff
             if m = /^index ([0-9a-f]{4,40})\.\.([0-9a-f]{4,40})( ......)*/.match(line)
               final[current_file][:src] = m[1]
               final[current_file][:dst] = m[2]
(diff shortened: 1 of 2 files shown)


jcouball:
While this might do for a temporary fix, I don't like this solution because while it filters most things out, it could still result in a false positive.

The real problem here is this line:
```ruby
git_cmd = "#{Git::Base.config.binary_path} #{global_opts} #{cmd} #{opts} #{command_opts[:redirect]} 2>&1"
```

Which forces STDERR to be captured along with STDOUT. This causes problems in a numerous places.

Needless to say, fixing this would take a lot more effort than your quick hack.

Let me consider this a bit more. As always, I am open to additional feedback.

aalong-tr:
Thank you for following up! I had figured that stderr was getting conflated with the rest of the output but hadn't found where the command was actually run, and wasn't certain to what extent the library relies on stderr in other scenarios.

I definitely agree that my solution is crude, and risks silently swallowing other types of problems; but at minimum, `process_full_diff` should probably offer a more graceful failure mode, as the current exception for unexpected output is very difficult to reason about. 

Thanks again, both for the extra context and for your contributions to open source.

jcouball:
@aalong-tr I am closing this in favor of #617 which will fix the underlying problem reported in #574. Happy to have your feedback there if you would like.

PR #589, opened 2022-08-02
Support the commit --no-gpg-sign flag

bradleybuda:
### Description

Add the ability to send the `--no-gpg-sign` option to git. This is the default behavior in a "vanilla" git install, but if gpg signing is configured elsewhere (i.e. in `~/.gitconfig`) then you can use this option to override and disable it for a particular commit.

Triggered by `gpg_sign: false` - see test and README updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,9 @@ g.commit('message', gpg_sign: true)
 key_id = '0A46826A'
 g.commit('message', gpg_sign: key_id)
 
+# Skip signing a commit (overriding any global gpgsign setting)
+g.commit('message', gpg_sign: false)
+
 g = Git.clone(repo, 'myrepo')
 g.chdir do
 new_file('test-file', 'blahblahblah')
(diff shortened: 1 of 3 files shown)


jcouball:
@bradleybuda I like this PR, but I request the following change:

Instead of passing `gpg_sign` as false, I think it would be more consistent with the rest of the API to pass `no_gpg_sign`. You can see examples of this for the following:
* `git commit --no-verify`
* `git merge --no-commit`
* `git merge --no-ff`
* `git revert --no-edit`

The design of the API mimics what you would pass on the command line. Like it or not, that is the design we have today.

jcouball:
I have fixed the Ruby 2.7 windows-latest build in #591

bradleybuda:
Thanks for the comment @jcouball, updated the patch. I hadn't noticed that there were some other `--no-`-prefixed flags already in the code so agree that this should be consistent.

jcouball (approved):
Looks great! Thanks for the thorough tests!

Issue #249, opened 2015-10-22
Create a new branch from a startpoint?


zhouguangming:
Looks like:  git.branch('branchname', start_point: 'a-commit-or-a-branch').create

jelaniwoods:
Is something like this possible with this gem?

PR #597, opened 2022-09-27
Add start_point option for checkout command

Vasfed:

### Description
This adds support for `start-point` git option when creating a new branch during checkout.
Fixes https://github.com/ruby-git/ruby-git/issues/249

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -764,11 +764,21 @@ module Git
       command('branch', '-D', branch)
     end
 
+    # Runs checkout command to checkout or create branch
+    #
+    # accepts options:
+    #  :new_branch
+    #  :force
+    #  :start_point
+    #
+    # @param [String] branch
+    # @param [Hash] opts
     def checkout(branch, opts = {})
       arr_opts = []
       arr_opts << '-b' if opts[:new_branch] || opts[:b]
       arr_opts << '--force' if opts[:force] || opts[:f]
       arr_opts << branch
+      arr_opts << opts[:start_point] if opts[:start_point] && arr_opts.include?('-b')
 
       command('checkout', arr_opts)
     end
--- a/tests/units/test_lib.rb
+++ b/tests/units/test_lib.rb
@@ -87,6 +87,19 @@ class TestLib < Test::Unit::TestCase
     assert(@lib.checkout('master'))
   end
 
+  def test_checkout_with_start_point
+    assert(@lib.reset(nil, hard: true)) # to get around worktree status on windows
+
+    actual_cmd = nil
+    @lib.define_singleton_method(:run_command) do |git_cmd, &block|
+      actual_cmd = git_cmd
+      super(git_cmd, &block)
+    end
+
+    assert(@lib.checkout('test_checkout_b2', {new_branch: true, start_point: 'master'}))
+    assert_match(%r/checkout ['"]-b['"] ['"]test_checkout_b2['"] ['"]master['"]/, actual_cmd)
+  end
+
   # takes parameters, returns array of appropriate commit objects
   # :count
   # :since
--- a/README.md
+++ b/README.md
@@ -265,6 +265,7 @@ g.branch('existing_branch').checkout
 g.branch('master').contains?('existing_branch')
 
 g.checkout('new_branch')
+g.checkout('new_branch', new_branch: true, start_point: 'master')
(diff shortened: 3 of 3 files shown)


lib/git/lib.rb
```diff
@@ -764,11 +764,21 @@ def branch_delete(branch)
...
+    #  :start_point
+    #
+    # @param [String] branch
+    # @param [Hash] opts
     def checkout(branch, opts = {})
       arr_opts = []
       arr_opts << '-b' if opts[:new_branch] || opts[:b]
       arr_opts << '--force' if opts[:force] || opts[:f]
       arr_opts << branch
+      arr_opts << opts[:start_point] if opts[:start_point] && (arr_opts.include?('-b') || arr_opts.include?('--force'))
```
jcouball: In my reading of https://git-scm.com/docs/git-checkout, I think that either `-b`, `-B`, or `--orphan` is required to use `<start-point>`. I don't think that `--force` without `-b` would be a valid use case for adding `<start-point>`.

I think this line should be:

```ruby
  arr_opts << opts[:start_point] if opts[:start_point] && arr_opts.include?('-b')
```

tests/units/test_lib.rb
```diff
@@ -83,6 +83,7 @@ def test_commit_with_no_verify
 
   def test_checkout
     assert(@lib.checkout('test_checkout_b',{:new_branch=>true}))
+    assert(@lib.checkout('test_checkout_b2', {new_branch: true, start_point: 'master'}))
```
jcouball: I know that this is duplicating the tests that are already here in `test_checkout`, but it doesn't really test what you are changing. Only that the command (whatever command was constructed -- even if it is wrong) succeeded.

jcouball: In addition to this test, can you add a separate test (something like `test_checkout_with_start_point`) that tests that the command is constructed correctly like this one:

https://github.com/bradleybuda/ruby-git/blob/master/tests/units/test_commit_with_gpg.rb#L14

jcouball (changes requested):
Hello @Vasfed! Thank you for this pull request.

I do have some requested changes in my review comments. In addition to those changes, can you also update the README.md to mention that this flag is supported?

Also, the build fails on Windows which you can see here:
https://github.com/ruby-git/ruby-git/actions/runs/3138166086/jobs/5154249814#step:4:41

```text
===============================================================================
Error: test_checkout(TestLib):
  Git::GitExecuteError: git "--git-dir=D:/tmp/git_test1664642165171/working/.git" "--work-tree=D:/tmp/git_test1664642165171/working" "-c" "core.quotePath=true" "-c" "color.ui=false" checkout "-b" "test_checkout_b2" "master"  2>&1:error: Your local changes to the following files would be overwritten by checkout:
  	colon_numbers.txt
  	example.txt
  	scott/newfile
  	scott/text.txt
  Please commit your changes or stash them before you switch branches.
  Aborting
D:/a/ruby-git/ruby-git/lib/git/lib.rb:1144:in `command'
D:/a/ruby-git/ruby-git/lib/git/lib.rb:783:in `checkout'
D:/a/ruby-git/ruby-git/tests/units/test_lib.rb:86:in `test_checkout'
     83:
     84:   def test_checkout
     85:     assert(@lib.checkout('test_checkout_b',{:new_branch=>true}))
  => 86:     assert(@lib.checkout('test_checkout_b2', {new_branch: true, start_point: 'master'}))
     87:     assert(@lib.checkout('.'))
     88:     assert(@lib.checkout('master'))
     89:   end
===============================================================================
```

lib/git/lib.rb
Vasfed: Hello James! Thank you for review.
You're correct, I was a bit mislead by `branch --force` https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---force

This and other remarks should now be fixed in force-pushed branch.

Do not have a windows machine at hand for debugging, but the error looks like it should relate to the test running order and not windows itself, if so - probably should be fixed by test separation too.

jcouball: Looks great! Thank you.

jcouball (dismissed):
Looks great! Thank you for the PR and the changes.

Vasfed:
Fixed tests on windows

jcouball (approved):
I had missed the build errors during my last review but did notice them before merging.

Thank you for proactively fixing the problem.

Issue #600, opened 2022-11-04
Git::Branch private function parse_name doesn't work as documented


mblythe86:
### Subject of the issue
The code comments right before `parse_name` say one thing, but the code doesn't do it.

### Your environment
* git 2.16.2
* ruby 2.3.1
* ruby-git 1.12.0

### Steps to reproduce
```ruby
require 'git'
git = Git.open(Dir.pwd)
['origin','origin/master','remotes/origin/master','origin/master/v2'].each do |name|
  puts "Testing #{name}"
  puts git.branch(name).remote
  puts git.branch(name).name
  puts "\n"
end
```

### Expected behaviour
Based on this comment in `branch.rb`:
```
      # Example:
      #   parse_name('master') #=> [nil, 'master']
      #   parse_name('origin/master') #=> ['origin', 'master']
      #   parse_name('remotes/origin/master') #=> ['origin', 'master']
      #   parse_name('origin/master/v2') #=> ['origin', 'master/v2'] 
```
I'd expect the above code to print this:
```
Testing origin
nil
master

Testing origin/master
origin
master

Testing remotes/origin/master
origin
master

Testing origin/master/v2
origin
master/v2
```

### Actual behaviour
The above code prints this:
```
Testing origin
nil
master

Testing origin/master
nil
origin/master

Testing remotes/origin/master
origin
master

Testing origin/master/v2
nil
origin/master/v2
```

### Recommendation
It looks like the regexp in `parse_name` is wrong.  It should include the slash after 'remotes' in the non-capturing group:
```ruby
        if name.match(/^(?:remotes\/)?([^\/]+)\/(.+)/)
```

Issue #599, opened 2022-11-04
Git::Branch function update_ref doesn't work as expected


mblythe86:
### Subject of the issue
Git::Branch#update_ref creates the wrong ref name, resulting in "refname is ambiguous" warnings.

### Your environment
* git 2.16.2
* ruby 2.3.1
* ruby-git 1.12.0

### Steps to reproduce
```ruby
require 'git'
git = Git.init
File.write('foo','rev 1')
git.add('foo')
git.commit('rev 1')
git.branch('testing').create
puts "HEAD", git.revparse('HEAD')
puts "testing", git.revparse('testing')
puts "refs/heads/testing", git.revparse('refs/heads/testing')
puts "\n\n"
File.write('foo','rev 2')
git.add('foo')
git.commit('rev 2')
git.branch('testing').update_ref(git.revparse('HEAD'))
puts "HEAD", git.revparse('HEAD')
puts "testing", git.revparse('testing')
puts "refs/heads/testing", git.revparse('refs/heads/testing')
```

### Expected behaviour
I'd expect `git.branch('testing').update_ref(<HASH>)` to behave the same as `git branch -f testing <HASH>` from the command line.

If that were the case, the code above should print the following (sha values may vary):
```
HEAD
2a0b5b7b...
testing
2a0b5b7b...
refs/heads/testing
2a0b5b7b...

HEAD
decf91cb...
testing
decf91cb...
refs/heads/testing
decf91cb...
```

### Actual behaviour
The code above prints this:
```
HEAD
2a0b5b7b...
testing
2a0b5b7b...
refs/heads/testing
2a0b5b7b...

HEAD
decf91cb...
testing
warning: refname 'testing' is ambiguous.
decf91cb...
refs/heads/testing
2a0b5b7b...
```

It looks like instead of updating the file `.git/refs/heads/testing` with the new hash, instead the call to `update_ref()` creates a new file `.git/testing` with the new hash.

### Discussion
For reference, this is the current implementation of Git::Branch#update_ref:

```ruby
def update_ref(commit)
  @base.lib.update_ref(@full, commit)
end
```

And this is the implementation of Git::Lib#update_ref that it's calling:

```ruby
def update_ref(branch, commit)
  command('update-ref', [branch, commit])
end
```

[...]

PR #601, opened 2022-11-04
Branch-related bugfixes

mblythe86:
Fixes both ruby-git#599 and ruby-git#600

Also fixes argument name of update_ref.  I'm assuming it's supposed to be analogous to the command line `git update-ref`, which doesn't directly use a branch name.

Signed-off-by: Matt Blythe <mblythester+git@gmail.com>

- [N/A] Ensure that your contributions contain documentation if applicable.

### Description
Fixes for #599 and #600

### DIFF
--- a/lib/git/branch.rb
+++ b/lib/git/branch.rb
@@ -78,7 +78,11 @@ module Git
     end
     
     def update_ref(commit)
-      @base.lib.update_ref(@full, commit)
+      if @remote
+        @base.lib.update_ref("refs/remotes/#{@remote.name}/#{@name}", commit)
+      else
+        @base.lib.update_ref("refs/heads/#{@name}", commit)
+      end
     end
     
     def to_a
@@ -114,7 +118,7 @@ module Git
       # param [String] name branch full name.
       # return [<Git::Remote,NilClass,String>] an Array containing the remote and branch names. 
       def parse_name(name)
-        if name.match(/^(?:remotes)?\/([^\/]+)\/(.+)/)
+        if name.match(/^(?:remotes\/)?([^\/]+)\/(.+)/)
           return [Git::Remote.new(@base, $1), $2]
         end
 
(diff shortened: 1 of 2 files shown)


jcouball:
The fix for #599 seems good but I'd like to see the following test added to `test/units/test_branch.rb`. This is a slightly simplified version of the problem you mentioned in #599.

```ruby
  def test_branch_update_ref
    in_temp_dir do |path|
      git = Git.init
      File.write('foo','rev 1')
      git.add('foo')
      git.commit('rev 1')
      git.branch('testing').create
      File.write('foo','rev 2')
      git.add('foo')
      git.commit('rev 2')
      git.branch('testing').update_ref(git.revparse('HEAD'))

      # Expect the call to Branch#update_ref to pass the full ref name for the
      # of the testing branch to Lib#update_ref
      assert_equal(git.revparse('HEAD'), git.revparse('refs/heads/testing'))
    end
  end
```

jcouball:
I think that the fix for #600 in `Git::Branch#parse_name` is something that I'd like to discuss further. 

First off, #600 fails to note the externally visible issue caused by the insufficient implementation of this private method. Having a failing test would be very helpful in discussing this problem.

`parse_name` is only used by `Git::Branch#initialize`.

`Git::Branch.new` is only called from:
* `Git::Branches#initialize`: the branch name is taken from `git branch -a`. e.g. `master` for a local branch or `remotes/{remote-name}/{branch-name}` for a remote branch
* `Git::Remote#branch`: the branch name is given as `{remote-name}/{branch-name}`
* `Git::Base#branch`: the branch name is only a name i.e. a local branch

The problem you ran into is easily duplicated for other cases that your fix does not solve.  For example, what happens when you create a (perfectly valid) local branch named `refs/remotes/origin/master`?

I am thinking that `Git::Branch#initialize` should probably be changed to have this signature: `def initialize(base, remote_name, branch_name)` to avoid this ambiguity. We would have to figure out how this impacts `Git::Base#branch` which is the only of these methods that should be used externally. Maybe document that it should only be used to create local branches (vs. remote) and have it add the "refs/heads/" prefix to the branch name before calling `Git::Branch.new`.

jcouball:
@mblythe86 I have added your change plus the test I suggested in #626.

PR #615, opened 2023-02-04
Simplify test execution

urbanautomaton:

### Description

Loading files for tests currently involves a lot of manually-constructed relative requires, as well as a `Dir.chdir` in the `tests/all_tests.rb` file which adjusts where requires need to be relative from.

Apart from some boilerplate, the main effect of this is that it's hard to run a single test file from the command line, as you need to edit the `all_tests.rb` file.

This makes things a bit simpler by:

* Adding the `tests` directory to the load path for the `rake test` task
* Removing the `chdir` from `all_tests.rb` to avboid a bunch of path complexity
* Replacing all relative requires with direct ones.

You can now run a single test file from the root directory of the project in a fairly standard way:

    $ ruby -Ilib:tests tests/units/test_object.rb

### Feedback

These changes seem simpler to me, but I haven't made a huge effort to investigate why the loading was the way it was. Are there any ways of running or writing tests that my changes would obstruct?

### DIFF
--- a/tests/all_tests.rb
+++ b/tests/all_tests.rb
@@ -1,8 +1,0 @@
-Dir.chdir(File.dirname(__FILE__)) do
-  Dir.glob('**/test_*.rb') do |test_case|
-    require "#{File.expand_path(File.dirname(__FILE__))}/#{test_case}"
-  end
-end
-
-# To run a single test:
-# require_relative 'units/test_lib_meets_required_version'
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,18 @@ In order to ensure high quality, all pull requests must meet these requirements:
   * The entire test suite must pass when `bundle exec rake default` is run from the
     project's local working copy.
 
+While working on specific features you can run individual test files or
+a group of tests using `bin/test`:
+
+    # run a single file:
+    $ bin/test tests/units/test_object.rb
+
+    # run multiple files:
+    $ bin/test tests/units/test_object.rb tests/units/test_archive.rb
+
+    # run all unit tests:
+    $ bin/test
+
 ### Continuous integration
   * All tests must pass in the project's [GitHub Continuous Integration build](https://github.com/ruby-git/ruby-git/actions?query=workflow%3ACI)
     before the pull request will be merged.
--- a/Rakefile
+++ b/Rakefile
@@ -1,16 +1,20 @@
 require 'bundler/gem_tasks'
 require 'English'
 
-require "#{File.expand_path(File.dirname(__FILE__))}/lib/git/version"
+require 'git/version'
 
 default_tasks = []
 
 desc 'Run Unit Tests'
 task :test do
-  sh 'git config --global user.email "git@example.com"' if `git config user.email`.empty?
-  sh 'git config --global user.name "GitExample"' if `git config user.name`.empty?
+  sh 'ruby bin/test'
 
-  require File.dirname(__FILE__) + '/tests/all_tests.rb'
+  # You can run individual test files (or multiple files) from the command
+  # line with:
+  #
+  # $ bin/test tests/units/test_archive.rb
+  #
+  # $ bin/test tests/units/test_archive.rb tests/units/test_object.rb
 end
 default_tasks << :test
 
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,19 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'bundler/setup'
+
+`git config --global user.email "git@example.com"` if `git config user.email`.empty?
+`git config --global user.name "GitExample"` if `git config user.name`.empty?
+
+project_root = File.expand_path(File.join(__dir__, '..'))
+
+$LOAD_PATH.unshift(File.join(project_root, 'tests'))
+
+if ARGV.empty?
+  paths = Dir.glob(File.join(project_root, 'tests/**/test_*.rb'))
+else
+  paths = ARGV.map { |p| File.join(project_root, p) }
+end
+
+paths.each { |p| require p }
--- a/tests/test_helper.rb
+++ b/tests/test_helper.rb
@@ -8,21 +8,13 @@ require "git"
 
 class Test::Unit::TestCase
 
-  def set_file_paths
-    cwd = FileUtils.pwd
-    if File.directory?(File.join(cwd, 'files'))
-      @test_dir = File.join(cwd, 'files')
-    elsif File.directory?(File.join(cwd, '..', 'files'))
-      @test_dir = File.join(cwd, '..', 'files')
-    elsif File.directory?(File.join(cwd, 'tests', 'files'))
-      @test_dir = File.join(cwd, 'tests', 'files')
-    end
+  TEST_ROOT = File.expand_path(__dir__)
+  TEST_FIXTURES = File.join(TEST_ROOT, 'files')
 
-    @wdir_dot = File.expand_path(File.join(@test_dir, 'working'))
-    @wbare = File.expand_path(File.join(@test_dir, 'working.git'))
-    @index = File.expand_path(File.join(@test_dir, 'index'))
+  BARE_REPO_PATH = File.join(TEST_FIXTURES, 'working.git')
 
-    @wdir = create_temp_repo(@wdir_dot)
+  def clone_working_repo
+    @wdir = create_temp_repo('working')
   end
 
   teardown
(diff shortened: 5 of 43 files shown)


urbanautomaton:
In fact, I got a bit carried away and did a more extensive bit of tidying among the tests. The [extra work is over here](https://github.com/ruby-git/ruby-git/compare/urbanautomaton:general-test-cleaning?w=1); there are a bunch of commits because I was working in small low-risk changes, but the summary is:

* **Overall goals:**
  * make individual tests easier to run
  * clarify use of fixture repo clones
* **Steps taken:**
  * Eliminate complexity due to regular changes of working directory
  * Make individual test files runnable via `ruby -Ilib:tests path/to/test_file.rb`
  * Extract constants to refer to fixture locations
  * Reduce unnecessary cloning of certain fixture repos
  * Begin using helper methods to clarify which tests are using which fixture repo

My hope is that once that last point is completed it should be much easier to then work out which fixture repo usage can be eliminated.

I haven't pushed all that to this PR yet because I'm conscious it's a lot to review and spare time is precious, but if any of that sounds of use, I'll be glad to include it.

tests/all_tests.rb
```diff
@@ -1,8 +1,9 @@
...
-    require "#{File.expand_path(File.dirname(__FILE__))}/#{test_case}"
-  end
+Dir.glob('tests/**/test_*.rb') do |test_case|
+  require File.expand_path(test_case)
 end
 
-# To run a single test:
-# require_relative 'units/test_lib_meets_required_version'
+# To run a single test, comment the above and uncomment:
+# require 'tests/units/test_lib_meets_required_version'
```
jcouball: I think we should REMOVE the recommendation in the comment above to edit this file to run a single test and just leave the comment below about running a single test from the command line.

tests/all_tests.rb
```diff
@@ -1,8 +1,9 @@
-Dir.chdir(File.dirname(__FILE__)) do
-  Dir.glob('**/test_*.rb') do |test_case|
-    require "#{File.expand_path(File.dirname(__FILE__))}/#{test_case}"
-  end
+Dir.glob('tests/**/test_*.rb') do |test_case|
+  require File.expand_path(test_case)
```
jcouball: How about this to avoid 39 calls to `File.expand_path`:

```ruby
require test_case[6..-1]
```

jcouball:
> These changes seem simpler to me, but I haven't made a huge effort to investigate why the loading was the way it was. Are there any ways of running or writing tests that my changes would obstruct?

Not that I know of. If there is, they are unknown to me. These changes look like a step in the right direction to me.

I haven't had the chance to look at your extra work yet.

jcouball:
@urbanautomaton I did want to get your feedback on an alternate idea to run individual tests. Is it a good idea or too hacky?

What if you could simply run a single test right from the rake command line:

```shell
$ rake test tests/units/test_archive.rb                             
Loaded suite /Users/james/.asdf/installs/ruby/3.2.0/bin/rake
Started
.
Finished in 0.273002 seconds.
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 15 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
3.66 tests/s, 54.94 assertions/s
$
```
or multiple tests:

```shell
rake test tests/units/test_archive.rb tests/units/test_worktree.rb      
Loaded suite /Users/james/.asdf/installs/ruby/3.2.0/bin/rake
Started
.....
Finished in 0.860063 seconds.
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
5 tests, 28 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
5.81 tests/s, 32.56 assertions/s
$
```

Here is the rake code that enables it:

```ruby
desc 'Run Unit Tests'
task :test do
  sh 'git config --global user.email "git@example.com"' if `git config user.email`.empty?
  sh 'git config --global user.name "GitExample"' if `git config user.name`.empty?

  $LOAD_PATH << 'tests'

  test_files = []
  if (test_index = ARGV.index('test'))
    (test_index + 1 .. ARGV.size - 1).each do |i|
      break unless ARGV[i] =~ /^tests\//
      test_files << ARGV[i][6..-1]
    end
  end
  test_files << 'all_tests' if test_files.empty?

  test_files.each { |test_file| require test_file }
end
```

tests/all_tests.rb
urbanautomaton: I was a bit unsure whether the intent of the array slicing would be clear to the reader; instead I've managed to get `Dir.glob` spitting out full absolute paths that can be directly `require`d without further manipulation. I've pushed an amended commit, see what you think?

urbanautomaton:
Thanks for the review!

On the rake modification: I know what you mean, it'd be nice to have a single entry point for running tests, and I did consider trying to pass the file args via rake.

The thing that makes me wary of using `ARGV` in rake tasks though is that it's kinda fighting against the way rake itself handles task arguments (i.e. the official rake way of passing in args would be `rake test[path/to/file_1.rb,path/to/file_2.rb]`). And in this case, it relies on `test` being the only task invoked, otherwise the `ARGV` offsets used would be wrong, so I think it's a bit fragile.

If we did want a single entry point capable of running all tests, one test or a subset of tests, I might be tempted to introduce a very simple `bin/test` executable to handle those use cases. This could become the documented way of running tests for contributors, and the rake task could remain for CI purposes, just invoking that executable to run all tests. What do you think?

tests/all_tests.rb
urbanautomaton: In fact `all_tests.rb` felt pointlessly small now so I've gotten rid of it and pushed the glob up to the rake task.

jcouball:
> And in this case, it relies on test being the only task invoked, otherwise the ARGV offsets used would be wrong, so I think it's a bit fragile.

The way I have implemented it allows other tasks to be passed on the command line like:

```text
rake build test tests/units/test_archive.rb tests/units/test_profile.rb yardstick
```

It does work against rake's built in method for passing args, but I find that method to be a bit cumbersome since the square brackets need escaping in zsh. 

Does that help?

jcouball:
Also, I reviewed the "extra work" you did and would love it if those commits were added to this PR. I will squash them all into one commit when I merge to master.

urbanautomaton:
Okay cool, I've pulled in those extra commits. 👍🏻 

> The way I have implemented it allows other tasks to be passed on the command line[...]

Ah yes, I'd missed that you're scanning forwards for the first `ARGV` entry that looks like a test file. And agreed that the square-bracket syntax is super cumbersome; I don't think it's a good fit here at all. 👍🏻

TBH my instinct is that we'd still be working around rake's official way of doing things though, which is likely to be a bit fragile and maybe cause developer confusion. So my leaning would remain towards having a separate test runner executable; [I've sketched out what that might look like here](https://github.com/ruby-git/ruby-git/commit/120b3ccb5d8a41c3d6871feb64fb16f6a358402e), see what you think!

jcouball:
> Okay cool, I've pulled in those extra commits. 👍🏻

Looks like you broke a test on Windows. Can you take a stab at fixing it? I am guessing since you didn't notice these broken tests, you aren't running windows. Neither do I... so I'd have to guess a solution and then commit the changes to see if the CI tests pass.

> TBH my instinct is that we'd still be working around rake's official way of doing things though, which is likely to be a bit fragile and maybe cause developer confusion. So my leaning would remain towards having a separate test runner executable; [I've sketched out what that might look like here](https://github.com/ruby-git/ruby-git/commit/120b3ccb5d8a41c3d6871feb64fb16f6a358402e), see what you think!

Yes, you convinced me.  Please add your changes to this PR. If documentation for how tests are run needs to be updated, please do that too. Maybe we should add a bullet in [this section of CONTRIBUTING.md](https://github.com/ruby-git/ruby-git/blob/master/CONTRIBUTING.md#unit-tests) that says something about how to run one or more specific test files.

urbanautomaton:
Okay cool - I fixed the windows jobs by backing out one of the smaller changes I made. I'm not sure of the exact cause but there's definitely something suspect going on in the tests that use multiple interlinked fixture repos, and I think that's best left to be investigated later.

Otherwise I've added the test runner executable and documented it in `CONTRIBUTING.md`, and squashed everything into an omnibus commit with a summary commit message. Ooh, and I added one final cheeky change which was to rename `#set_file_paths` to `#clone_working_repo` because that's all that method does now.

I'm definitely stopping there for the moment though; from my end it's ready to go. 😅

jcouball (approved):
I have reviewed all changes. These are welcome changes. Nice job!

jcouball:
This PR:

* Reduced lines of code for code in the tests directory tree by 5% (from 2,700 to 2,555 lines of code)
* Reduced the time it takes to run the entire test suite (on my machine) by 12% (from 27.5 seconds to 24.4 seconds)
* Made it easier to run a single test file or multiple test files so contributors will have less to figure out.
* In many cases, simplified tests so they are easier to understand (and duplicate if needed).

On the whole, this PR removed a lot of tech debt and made this gem easier to change.

Great job Simon (@urbanautomaton )! This is a great contribution!

PR #622, opened 2023-02-21
Refactor error thrown when a git command fails

jcouball:

### Description
In preparation for refactoring how Git commands are executed by this gem (see #617), add two new classes: `Git::FailedError` and `Git::SignaledError` to indicate different failure conditions when running the Git CLI.

* `Git::FailedError` is raised when the `git` command returns a non-zero exitstatus.
* `Git::Signaled` is raised when the `git` command fails because of an uncaught signal (like a kill signal).

Both of these new errors derive from `Git::GitExecuteError` so existing code that was rescuing this error will still work as before.

Both of these new errors expect to be initialized with an instance of another new class added in this PR: `Git::CommandLineResult`. Objects of this class capture:

* the git command executed
* the resulting Process::Status
* the command's stdout
* the command's stderr

The intent is to give callers better context as to the type of error (non-zero exitstatus vs. uncaught signal) as well as provide better context as to the command executed, the ending status of the process (including exitstatus or signal #), and the captured command line output.

`Git::FailedError` has been fully integrated into `Git::Lib#command`.  Any situation where a `Git::GitExecuteError` would have been raised before now raises a `Git::FailedError`. As mentioned before, since `Git::FailedError` is a subclass of `Git::GitExecuteError`, no change is required to client code.

`Git::SignaledError` has not yet be integrated into `Git::Lib#command`.


lib/git/failed_error.rb
```diff
@@ -0,0 +1,51 @@
...
+    #   result = Git::CommandLineResult.new(%w[git status], $?, '', "failed")
+    #   error = Git::FailedError.new(result)
+    #   error.message #=>
+    #     "[\"git\", \"status\"]\nstatus: pid 89784 exit 1\nstderr: \"failed\""
+    #
+    # @param result [Git::CommandLineResult] the result of the git command including
+    #   the git command, status, stdout, and stderr
+    #
+    def initialize(result)
+      super("#{result.git_cmd}\nstatus: #{result.status}\nstderr: #{result.stderr.inspect}")
```
alexjfisher, line 28: @jcouball I think stdout also has to be printed here since stderr is redirected to stdout?? https://github.com/ruby-git/ruby-git/blob/5b0e1c8b62188e2a80427bba3f625aaaf84e7350/lib/git/lib.rb#L1194

Issue #516, opened 2021-02-25
Git.ls_remote fails when stderr isn't empty


serhii-shnurenko:
### Subject of the issue
When the output of `ls-remote` command is parsed both stout and stderr parsed, what causes the script to fail because of the unexpected git command output.

### Your environment
* git version 1.7.1, ruby-git 1.5.0
* ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]

### Expected behaviour
Git.ls_remote working even if something fails into stderr.
Output to stderr parsed separately from stdout.
IDK how complicated is it, but why not switch `Open3::popen3` as command runner?

### Steps to reproduce
You should have output in stderr for your git command. Then `Git.ls_remote` will fail.

### Actual behaviour
When running Git.ls_remote and stderr isn't empty, the parsing of ls-remote command will fail because warning in the output causes:
```
NoMethodError: undefined method `split' for nil:NilClass
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git/lib.rb:419:in `block (2 levels) in ls_remote'
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git/lib.rb:417:in `each'
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git/lib.rb:417:in `block in ls_remote'
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git/lib.rb:416:in `tap'
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git/lib.rb:416:in `ls_remote'
  /opt/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/git-1.5.0/lib/git.rb:149:in `ls_remote'
....
```
The reason is that git commands on this host for some reason output warnings about hostkey everytime. And I still searching for the reasons, why `~/.ssh/known_hosts` file in the home is ignored, but I believe that this kind of warnings shouldn't fail program execution.
When I running my script, I'm getting:
```
RSA host key for IP address '**********' not in list of known hosts.
```
Which then not splitted by the tabs because there are no tabs in this string
[...]

Issue #495, opened 2020-12-01
Commit message escaping issues under Windows


sergio-bobillier:
### Subject of the issue
The current escape method for Windows is basically not doing anything:
https://github.com/ruby-git/ruby-git/blob/f9abb180b84eddbc31ef578d4587b8612455afa3/lib/git/lib.rb#L1156

```ruby
    def escape_for_windows(s)
      # Windows does not need single quote escaping inside double quotes
      %Q{"#{s}"}
    end
```

This works fine for single quotation marks but if the message contains a double quote the commit method breaks:

```text
I, [2020-12-01T13:47:55.042634 #5488]  INFO -- : git "--git-dir=Z:/Trash/repo/.git" "--work-tree=Z:/Trash/repo" "-c" "color.ui=false" commit "--message="Hello World""  2>&1
D, [2020-12-01T13:47:55.042873 #5488] DEBUG -- : error: pathspec 'World' did not match any file(s) known to git
Git::GitExecuteError: git "--git-dir=Z:/Trash/repo/.git" "--work-tree=Z:/Trash/repo" "-c" "color.ui=false" commit "--message="Hello World""  2>&1:error: pathspec 'World' did not match any file(s) known to git
from C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/git-1.7.0/lib/git/lib.rb:989:in `command'
```

### Your environment
* ruby-git: 1.7.0
* Ruby:  2.6.4
* Windows 10

### Steps to reproduce

Create a repository, open with with Git and then create a commit with a newline on the message:

```
mkdir repo
cd repo
git init
echo "Hello World" > hello.txt
git add hello.txt
pry
require 'git'
git = Git.open('.')
git.commit('"Hello World"')
```

### Expected behaviour
I would expect this to work like it does on Unix-like systems. I.e. the commit is performed and the message includes the quotation marks.

### Actual behaviour
A `Git::GitExecuteError` is raised and the origin of the problem is the incorrectly escaped commit message.

jcouball:
Fixed with #684 and released in [git-2.0.0-pre1](https://rubygems.org/gems/git/versions/2.0.0.pre1)

Issue #441, opened 2020-01-31
Streaming stdout instead of one response


aquaflamingo:
### Subject of the issue
Hi there, :wave: great library. 

For long running git operations, I'm wonder the thoughts on adding a "stream stdout" option to the library instead of the [current](https://github.com/ruby-git/ruby-git/blob/f5cd6a62d3d5b9f716d1814eb74159bf2679b25c/lib/git/lib.rb#L963) implementation which takes the bulk response back stores it in a variable and returns it.

Is this something you think would be relatively simple and / or desirable?

jcouball:
#617 should enable doing this. It isn't possible with the current way that the library calls `git` with backticks.

jcouball:
Fixed with #684 and released in [git-2.0.0-pre1](https://rubygems.org/gems/git/versions/2.0.0.pre1)

PR #673, opened 2023-10-20
Remove calls to Dir.chdir

fxposter:

### Description
In multithreaded environment Dir.chdir changes current directory of all process' threads, so other threads might misbehave.

Base#chdir, Base#with_working and Base#with_temp_working were left intact, cause they are not used internally, so it's an explicit user decision to use them.

Fixes https://github.com/ruby-git/ruby-git/issues/355

### DIFF
--- a/lib/git.rb
+++ b/lib/git.rb
@@ -244,7 +244,7 @@ module Git
     options.delete(:remote)
     repo = clone(repository, name, {:depth => 1}.merge(options))
     repo.checkout("origin/#{options[:branch]}") if options[:branch]
-    Dir.chdir(repo.dir.to_s) { FileUtils.rm_r '.git' }
+    FileUtils.rm_r File.join(repo.dir.to_s, '.git')
   end
 
   # Same as g.config, but forces it to be at the global level
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -1,5 +1,6 @@
 require 'git/base/factory'
 require 'logger'
+require 'open3'
 
 module Git
   # Git::Base is the main public interface for interacting with Git commands.
@@ -66,11 +67,11 @@ module Git
     def self.root_of_worktree(working_dir)
       result = working_dir
       status = nil
-      Dir.chdir(working_dir) do
-        git_cmd = "#{Git::Base.config.binary_path} -c core.quotePath=true -c color.ui=false rev-parse --show-toplevel 2>&1"
-        result = `#{git_cmd}`.chomp
-        status = $?
-      end
+
+      git_cmd = "#{Git::Base.config.binary_path} -c core.quotePath=true -c color.ui=false rev-parse --show-toplevel 2>&1"
+      result, status = Open3.capture2(git_cmd, chdir: File.expand_path(working_dir))
+      result = result.chomp
+
       raise ArgumentError, "'#{working_dir}' is not in a git working tree" unless status.success?
       result
     end
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -2,6 +2,7 @@ require 'git/failed_error'
 require 'logger'
 require 'tempfile'
 require 'zlib'
+require 'open3'
 
 module Git
   class Lib
@@ -441,7 +442,10 @@ module Git
     def list_files(ref_dir)
       dir = File.join(@git_dir, 'refs', ref_dir)
       files = []
-      Dir.chdir(dir) { files = Dir.glob('**/*').select { |f| File.file?(f) } } rescue nil
+      begin
+        files = Dir.glob('**/*', base: dir).select { |f| File.file?(File.join(dir, f)) }
+      rescue
+      end
       files
     end
 
@@ -579,15 +583,7 @@ module Git
     end
 
     def config_get(name)
-      do_get = Proc.new do |path|
-        command('config', '--get', name)
-      end
-
-      if @git_dir
-        Dir.chdir(@git_dir, &do_get)
-      else
-        do_get.call
-      end
+      command('config', '--get', name, chdir: @git_dir)
     end
 
     def global_config_get(name)
@@ -595,15 +591,7 @@ module Git
     end
 
     def config_list
-      build_list = Proc.new do |path|
-        parse_config_list command_lines('config', '--list')
-      end
-
-      if @git_dir
-        Dir.chdir(@git_dir, &build_list)
-      else
-        build_list.call
-      end
+      parse_config_list command_lines('config', '--list', chdir: @git_dir)
     end
 
     def global_config_list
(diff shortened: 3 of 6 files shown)


fxposter:
cc @jcouball

jcouball:
@fxposter thank you very much for this PR and #675 

This PR is still failing for JRuby on Windows. I faced the same problems when trying to solve these problems and more (#355, #363, #441, #495, #516, #574).

I almost have that solution over the finish line with #617. This is (to the best I can tell) only missing one JRuby issue which is tracked in ruby/jruby#7515.

I am wondering if you gave #617 a try? I would love feedback on that one.

fxposter:
I'll take a look, but do you really believe that you need to have "process execution" abstraction instead of using the existing ones? :)

PS. I'll get to a windows machine on the weekend, will try to check what is going on there, cause open3 specifically have code that should handle chdir/etc: https://github.com/ruby/open3/blob/master/lib/open3/jruby_windows.rb#L81

fxposter:
@jcouball can you review and run tests again?
I have found what was wrong with windows - due to the fact that tests use Dir.chdir themselves - when we passed chdir to open3 it didn't "inherit" it from the current Dir.chdir call that was done by tests:

```
# imagine that we run this script in dir /x

Dir.chdir('/y') {
  Dir.chdir('.') {
    Open3.capture2('ls') # will show files in /y
  }
}

Dir.chdir('/y') {
  Open3.capture2('ls', :chdir => '.') # will show files in /x
}
```

to avoid that I do `:chdir => File.expand_path('.')` which properly expands relative paths.

fxposter:
Also, before merging: I rewrote code like
```
Dir.chdir(dir) { files = Dir.glob('**/*').select { |f| File.file?(f) } }
```
to
```
files = Dir.glob('**/*', :base => dir).select { |f| File.file?(f) }
```
The end result is the same, but `:base` is only supported since ruby 2.5. If that is a problem - I can rewrite the code without using `:base` keyword argument.

jcouball:
Thanks @fxposter, we are getting very close! Only one error left in the Windows JRuby build.

I see that the failure is [in `test_tree_ops.rb` in a `with_index` block](https://github.com/ruby-git/ruby-git/blob/8481f8c6ef0dedd89374c85ca94751f5452bc414/tests/units/test_tree_ops.rb#L82-L94).

My first comment is that I am not a fan of this tests in this file since it is just one big test case with a lot of state changing and asserts sprinkled through the method. An isolated, minimal, failing test case would be ideal but not necessary.

The first think I would try is to use [`Git::Base#with_temp_index`](https://github.com/ruby-git/ruby-git/blob/master/lib/git/base.rb#L536-L546) instead of `Git::Base#with_index`. I see that there is code in that method to workaround some problem in JRuby when using `Tempfile` exactly the way the test does.

I honestly don't understand why the `with_index` block is being used in the first place. Maybe to use a "throw away" index?

Beyond that, maybe something related to the changes you are making might stick out to you.

I have changed the configuration of the GitHub actions for the CI build so that the build will run when you push changes. This way you won't have to wait for me or another contributor for feedback.

fxposter:
:(

I hoped that the last change fixes everything on jruby on windows. The tricky part here is that locally some tests are always failing for me on both Mac and Windows (while they pass in CI), so I assumed I fixed everything.

I'll try to focus on the failing one, as I have access to windows machine during the weekend.

jcouball:
Regarding your comments about which Ruby release should be supported, here is release information for some Ruby releases from the [Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/) page:

```
Ruby 3.0
status: security maintenance
release date: 2020-12-25
normal maintenance until: 2023-04-01
EOL: 2024-03-31 (expected)

Ruby 2.7
status: eol
release date: 2019-12-25
normal maintenance until: 2022-04-01
EOL: 2023-03-31

Ruby 2.6
status: eol
release date: 2018-12-25
normal maintenance until: 2021-04-01
EOL: 2022-04-12

Ruby 2.5
status: eol
release date: 2017-12-25
normal maintenance until: 2020-04-01
EOL: 2021-04-05
```

I think for this PR we could change [the required version of Ruby in the gemspec](https://github.com/ruby-git/ruby-git/blob/8481f8c6ef0dedd89374c85ca94751f5452bc414/git.gemspec#L25) to 2.5.

I notice that the CI build only goes back to 2.7 so there should probably be a build added for MRI 2.5 on Linux at least.

It would be my intention to follow up with an explicit policy that this gem follows for supported Ruby version in a separate PR. The policy could be something like "Support all non-EOL Ruby minor versions" (currently 3.0, 3.1, and 3.2). We could possibly include the latest EOL minor version as well (which would add 2.7 until 3.0 is EOL'd in March 2024).

jcouball:
> The tricky part here is that locally some tests are always failing for me on both Mac and Windows (while they pass in CI)

Huh, that shouldn't be and is concerning. If the CI builds pass and the local build works for me (which it does), they I'd accept the pull request. If you want to share details of your failure(s) we can try to debug. I am on Mac so that is where I can be most helpful.

fxposter:
sure. this is what I get on my mac by default (probably it's related to my user git's config, but I didn't try to debug that yet, since it failed on master as well):
```
 $ rake
ruby bin/test
Loaded suite bin/test
Started
E
=====================================================================================================================================================================================================================================================
Error: test_push(TestRemotes):
  Git::FailedError: git '--git-dir=/private/var/folders/m3/lnm6c86s2m99_bx5n8c4j_ww0000gn/T/d20231118-67431-4pwmvb/local/.git' '--work-tree=/private/var/folders/m3/lnm6c86s2m99_bx5n8c4j_ww0000gn/T/d20231118-67431-4pwmvb/local' '-c' 'core.quotePath=true' '-c' 'color.ui=false' 'push' 'testrem'  2>&1
  status: pid 70744 exit 128
  output: "fatal: You are pushing to remote 'testrem', which is not the upstream of\nyour current branch 'master', without telling me what to push\nto update which remote branch.\n"
/Users/pavelf/Projects/github/ruby-git/lib/git/lib.rb:1234:in `command'
/Users/pavelf/Projects/github/ruby-git/lib/git/lib.rb:996:in `push'
/Users/pavelf/Projects/github/ruby-git/lib/git/base.rb:396:in `push'
/Users/pavelf/Projects/github/ruby-git/tests/units/test_remotes.rb:226:in `block in test_push'
     223:       assert(!rem.status['test-file1'])
     224:       assert(!rem.status['test-file3'])
     225:
  => 226:       loc.push('testrem')
     227:
     228:       assert(rem.status['test-file1'])
     229:       assert(!rem.status['test-file3'])
/Users/pavelf/Projects/github/ruby-git/tests/test_helper.rb:68:in `block in in_temp_dir'
/Users/pavelf/Projects/github/ruby-git/tests/test_helper.rb:67:in `chdir'
/Users/pavelf/Projects/github/ruby-git/tests/test_helper.rb:67:in `in_temp_dir'
/Users/pavelf/Projects/github/ruby-git/tests/units/test_remotes.rb:205:in `test_push'
```

jcouball:
@fxposter what is the output of

```
git --version && git config --list --show-origin
```

jcouball:
Looks like you might have `push.default` set to `upstream`. I'll prepare a PR to explicitly set the `push.default` for this test.

fxposter:
```
color.ui=auto
push.default=tracking
push.autosetupremote=true
branch.autosetuprebase=always
init.defaultbranch=master
```
and lots of others, probably unrelated

jcouball:
It looks like `push.default=tracking` gives the same result as `push.default=upstream`.

jcouball:
@fxposter if you rebase with master, test_push(TestRemotes) should pass now.

PR #674, opened 2023-10-20
Fix commit tree with multiple parents

maths22:

### Description
This change fixes commit_tree to actually pass through multiple parents
correctly, while preserving backwards-compatibility by still handling
single commits passed into the `parents` field.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -1040,10 +1040,19 @@ module Git
       t.write(opts[:message])
       t.close
 
+      # Adapted from activesupport Array.wrap
+      parents = if opts[:parents].nil?
+        []
+      elsif opts[:parents].respond_to?(:to_ary)
+        opts[:parents].to_ary || [opts[:parents]]
+      else
+        [opts[:parents]]
+      end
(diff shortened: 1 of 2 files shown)


lib/git/lib.rb
```diff
@@ -1040,10 +1040,19 @@ def commit_tree(tree, opts = {})
       t.write(opts[:message])
       t.close
 
+      # Adapted from activesupport Array.wrap
+      parents = if opts[:parents].nil?
```
jcouball, line 1044: I read up on `Array.wrap` but don't understand why you would choose that implementation instead of `Kernel#Array`.  I am interested to hear why you would go this route instead of reducing this change to:

```Ruby
arr_opts += Array(opts[:parent]).map { |p| ['-p', p] }.flatten
```

I think calling `#to_a` if `#to_ary` doesn't exist would be the desired behavior.

jcouball:
@maths22 thank you for this contribution. After looking into the test_tree_ops.rb test case, I could see that these tests needed refactoring. This refactoring was done in #679 along with fixing the issue you submitted this fix for.

Apologies for taking so long.

PR #675, opened 2023-10-25
Do not change global env

fxposter:

### Description
Do not change ENV variables of a current process. Instead pass relevant environment to the actual shell command being run.

this PR is built on top of #673

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -69,8 +69,10 @@ module Git
       status = nil
 
       git_cmd = "#{Git::Base.config.binary_path} -c core.quotePath=true -c color.ui=false rev-parse --show-toplevel 2>&1"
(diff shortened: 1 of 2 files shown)


maths22:
Some version of this would be nice to have: the current global env version in practice means only one git operation can run at a time per process, which is rather inconvenient.  Unless I'm missing something, based on what was merged in  #673 I think this can be changed to just have "option 2" (capture2/popen2) and is then probably in a good state?

jcouball:
@maths22 

All methods I have tried did not work for one reason or another in JRuby on Windows.

I have released git-2.0.0-pre1 which I believe will address this issue in a different way. You can see PR #684 for details. This version still does not support JRuby for Windows, but I believe there is a path to victory there.

The API of this gem is unchanged. All the changes are below the hood.

jcouball:
Fixed in #684 and released in 2.0.0.

Issue #657, opened 2023-03-09
Implement a way to mark a public API as deprecated


jcouball:
Have the ability for the developer to indicate deprecated methods.

Calling deprecated methods (by default) should output a STDERR warning message saying that the deprecated method was called and what line of code called it.

The user of the gem should have the ability to suppress the warning messages or to raise an error when a deprecated method is called.

jcouball:
Fixed in #709

PR #709, opened 2024-05-05
Refactor errors that are raised by this gem

jcouball:
Refactor the errors raised by this gem into this hierarchy of classes defined in the `Git` module:

```text
Error
├─> CommandLineError
│   ├─> FailedError
│   └─> SignaledError
│       └─> TimeoutError
├─> ProcessIOError
└─> UnexpectedResultError
```

Other standard errors may also be raised like `ArgumentError`. Each method should document the errors it may raise.

Description of each Error class:

* `Error`: This catch-all error serves as the base class for other custom errors in this
  gem. This gem only raises errors that are a subclass of `Git::Error`.
* `CommandLineError`: This error is raised when there's a problem executing the git
  command line. This gem will raise a more specific error depending on how the
  command line failed.
* `FailedError`: This error is raised when the git command line exits with a non-zero
  status code that is not expected by the git gem.
* `SignaledError`: This error is raised when the git command line is terminated as a
  result of receiving a signal. This could happen if the process is forcibly
  terminated or if there is a serious system error.
* `TimeoutError`: This is a specific type of `SignaledError` that is raised when the
  git command line operation times out and is killed via the SIGKILL signal. This
  happens if the operation takes longer than the timeout duration configured in
  `Git.config.timeout` or via the `:timeout` parameter given in git methods that
  support this parameter.
* `ProcessIOError`: An error was encountered reading or writing to a subprocess.
* `UnexpectedResultError`: The command line ran without error but did not return
  the expected results.

For backward compatability, `Git::GitExecuteError` remains as an alias for `Git::Error`. It is considered
deprecated as of v2.0.0 and will be remove in v3.0.0.


jcouball:
Incidentally, this PR fixes #657 as activesupport was integrated in order to mark the original error class as deprecated.

PR #723, opened 2024-05-24
Support `no_color: true` on diff commands

nevinera:
We ran into an issue when combining the diff output from this gem with several of the diff-parsers, that only turned up for users that had `color.diff=always` set. Most likely because they liked still having diff colors after piping their diffs through grep/etc, but for _our_ purposes (the `quiet_quality` gem, which uses the fairly old `git_diff_parser` gem), it just broke change-detection whenever that setting was enabled, because the color escapes break the parsing.

This change allows us to supply an extra option `no_color: true` (it defaults to the current behavior) like this:

    git.diff(no_color: true).patch
    git.diff(base, sha, no_color: true).patch

and get un-colorized output more appropriate for parsing, even if the git-configuration has `color.diff=always`

Note - I looked for tests on the equivalent existing feature (the `color.ui` override) and didn't find any, but if you'd like _this_ feature to have tests, I'm happy to add some. I just wasn't sure what layer to test at - I've pretty much only worked in rspec thus far, and could use some guidance :-) (I'd normally test this sort of thing by mocking the Git::CommandLine instantiation, but that doesn't look like that preferred approach here)

Likewise, I think the `Git::Object::AbstractObject#diff` method qualifies as a 'public' method (it is mentioned in the readme examples), but none of the methods in that file had yard documentation yet, and I wasn't sure if I should start adding it for just the one method? Happy to iterate as much as is helpful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,11 +281,12 @@ g.describe('HEAD', {:all => true, :tags => true})
 g.diff(commit1, commit2).size
 g.diff(commit1, commit2).stats
 g.diff(commit1, commit2).name_status
+g.diff(commit1, commit2, no_color: true).patch
 g.gtree('v2.5').diff('v2.6').insertions
 g.diff('gitsearch1', 'v2.5').path('lib/')
 g.diff('gitsearch1', @git.gtree('v2.5'))
 g.diff('gitsearch1', 'v2.5').path('docs/').patch
-g.gtree('v2.5').diff('v2.6').patch
+g.gtree('v2.5').diff('v2.6', no_color: true).patch
 
 g.gtree('v2.5').diff('v2.6').each do |file_diff|
   puts file_diff.path
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -675,9 +675,10 @@ module Git
       Git::Object::Commit.new(self, self.lib.commit_tree(tree, opts))
     end
 
+    # @param [Boolean] no_color Prevent git from colorizing diff output
     # @return [Git::Diff] a Git::Diff object
-    def diff(objectish = 'HEAD', obj2 = nil)
-      Git::Diff.new(self, objectish, obj2)
+    def diff(objectish = 'HEAD', obj2 = nil, no_color: false)
+      Git::Diff.new(self, objectish, obj2, no_color: no_color)
     end
 
     # @return [Git::Object] a Git object
--- a/lib/git/diff.rb
+++ b/lib/git/diff.rb
@@ -4,10 +4,11 @@ module Git
   class Diff
     include Enumerable
 
-    def initialize(base, from = nil, to = nil)
+    def initialize(base, from = nil, to = nil, no_color: nil)
       @base = base
       @from = from && from.to_s
       @to = to && to.to_s
+      @no_color = no_color
 
       @path = nil
       @full_diff = nil
@@ -101,7 +102,7 @@ module Git
     private
 
       def cache_full
-        @full_diff ||= @base.lib.diff_full(@from, @to, {:path_limiter => @path})
+        @full_diff ||= @base.lib.diff_full(@from, @to, {:path_limiter => @path, :no_color => @no_color})
       end
 
       def process_full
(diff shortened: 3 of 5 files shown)


nevinera:
> I wonder if the default should be to turn off colorization for diffs in the first place. This would require, of course, another major version bump, but I think this is more inline (IMHO) with the intent of this gem to be a programmatic interface vs. a user interface. To be honest, that intent is hard to intuit since it isn't really well documented.
> 
> I am interested to hear your thoughts on this.
> 
> Coincidentally, I had recently written [Design Philosophy of the git Gem](https://gist.github.com/jcouball/f259d9c63a0cd49340e7696544a45b69) to give this guidance. I have just published in case you care and have any feedback.

My _gut_ says you're right, but I can't say with confidence that I know all the reason's people might set `color.diff` to "always" - the only reason I've seen people do it is so they can get colors in the output after piping git-diff to `grep`, so it seems like a safe guess to me?

But I don't think I would bother cranking out a major version for that if I were you - maybe go with the optional version for now, and add it to a list for when you have some more significant changes you want to make? The other thing I was going to suggest was that you _might_ prefer if I handled it as an options hash instead of a keyword argument - it seems like there might be other git-diff flags people'd like to be able to pass through at some point, and then when/if you **do** switch to automatically using `--no-color` you can just ignore the option when supplied, instead of introducing incompatibility.

> Thank you! I am supporting this gem alone and I value any chance for outside feedback.

It's been very useful to me, I'll see if there's anything in the tracker I can help with.

jcouball:
Hey, Eric! Thank you for your time and for caring about this gem. I really do appreciate it.

As I have studied this, I am leaning to globally disabling all color output.

I think an argument can be made that this was the original intent since we turn off color in general with `-c color.ui=false` and turn off color for `git log` by passing the `--no-color` option there.

The bummer is that I don't see any way to turn off color globally other than setting a bunch of configuration variables on the command line.

This includes setting:
* color.ui=false
* color.advice=false
* color.branch=false (override with --color option)
* color.diff=false (override with --color option)
* color.grep=false (override with --color option)
* color.push=false
* color.remote=false
* color.showBranch=false (override with --color option)
* color.status=false
* color.transport=false

Or maybe instead of always setting all these variables, just figure out which ones need to be set. I suspect that might not be worth the effort.

the command line would look something like this:

```
git -c color.ui=false \
  -c color.advice=false \
  -c color.branch=false \
  -c color.diff=false \
  -c color.grep=false \
  -c color.push=false \
  -c color.remote=false \
  -c color.showBranch=false \
  -c color.status=false \
  -c color.transport=false \
  [OTHER_GLOBAL_OPTS] COMMAND [COMMAND_OPTS] [COMMAND_ARGS]
```

And after thinking more about it, I think this warrants only a minor release instead of a major release. The rational is that this was the original intent all along.

Would love to hear your thoughts on this plan.

jcouball:
Eric, BTW wanted to say, I really love the idea of the quiet_quality gem. I have not used it myself, but I can see how it can be very helpful to teams that want to get their project on a better footing without incurring more tech debt.

Most of my other projects insist on 100% compliance for Rubocop, unit tests, and code coverage so quiet_quality would not be helpful there (I think).

One place where it might be useful is this gem. I'll put this down on my list of things to investigate (which is a long list unfortunately).

Very cool!

jcouball:
Ok, thought of one more alternative to turning off colorization for everything:

Always add the `-c color.ui=false` option to the command line and then only add specific color disabling options based on the command. For example, only set `-c color.status=false` when running `git status`. It would be the responsibility of each command to add this right option.

Also, Git::Lib#command and Git::Lib#command_lines would have to be extended to add a way to pass in what I called OTHER_GLOBAL_OPTIONS before.

nevinera:
You've convinced me - I think that if we're already applying `color.ui`, also supplyinh the others is just a matter of being more thorough, not a philosophical departure :-)

I'll open a different PR, and close this one when it's ready.

nevinera:
You've convinced me - I think that if we're already applying `color.ui`, also supplyinh the others is just a matter of being more thorough, not a philosophical departure :-)

I'll open a different PR, and close this one when it's ready.

Oh, and QQ is still mildly useful if you're fully in compliance - it does give a nice cli for running all of the tools concurrently, and can annotate PRs, instead of just failing builds. I could set you up an example PR if you're interested!

nevinera:
Here's [that PR](https://github.com/ruby-git/ruby-git/pull/728) - checked it against QQ, and it also resolves my issue there :-)

PR #724, opened 2024-05-26
When core.ignorecase, check Status for files case-insensitively

### LINKED ISSUE #586: `Status#changed?` is case-sensitive despite git config
### Subject of the issue
When the git repo is configured to track files case-insensitively and a file name has a different letter case, `Status#changed?` will not report if the file has been modified.

### Your environment
* git (1.11.0)
* ruby 2.7.3

### Steps to reproduce
- init a new git repo
- (have default config `core.ignoreCase true`)
- commit `FILE`
- rename to `file`
- edit `file`
- in ruby-git, call `Status#changed?("file")`

### Expected behaviour
expected to compare tracked file case-insensitively with given filename (ie `"FILE".downcase == "file".downcase`)

### Actual behaviour
returns false

nevinera:
Fixes #586 by making `Status#changed?` case-insensitive when `core.ignoreCase` is set. Also solve the matching unreported issues for `Status#added`, `Status#deleted`, and `Status#untracked`.

I did run into a bit of an issue trying to write thorough tests for these, because it turns out that _git itself_ doesn't support setting `core.ignoreCase` to a value that doesn't match the behavior of the file-system; [that option](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreignoreCase) is provided so you can tell git how the file-system is going to act, not so you can control git's behavior! In the tests here, I do break that guideline, but because I don't actually rename any files, `git` doesn't have trouble - just don't try to make it test the case that was actually _reported_, because that will not work.

### DIFF
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -108,7 +108,11 @@ module Git
     #     untracked?('lib/git.rb')
     # @return [Boolean]
     def untracked?(file)
-      untracked.member?(file)
+      if ignore_case?
+        untracked.keys.map(&:downcase).include?(file.downcase)
+      else
+        untracked.member?(file)
+      end
     end
 
     def pretty
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -34,7 +34,11 @@ module Git
     #     changed?('lib/git.rb')
     # @return [Boolean]
     def changed?(file)
-      changed.member?(file)
+      if ignore_case?
+        changed.keys.map(&:downcase).include?(file.downcase)
+      else
+        changed.member?(file)
+      end
     end
 
     # Returns an Enumerable containing files that have been added.
@@ -263,6 +267,16 @@ module Git
           @files[path] ? @files[path].merge!(data) : @files[path] = data
         end
       end
+    end
+
+    # It's worth noting that (like git itself) this gem will not behave well if
+    # ignoreCase is set inconsistently with the file-system itself. For details:
+    # https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreignoreCase
+    def ignore_case?
+      return @_ignore_case if defined?(@_ignore_case)
+      @_ignore_case = @base.config('core.ignoreCase') == 'true'
+    rescue Git::FailedError
+      @_ignore_case = false
     end
   end
 end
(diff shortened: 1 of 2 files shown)


nevinera:
This test passed easily locally (in the first commit), but something is going on in the CI context that is confusing me greatly - if you compare the output I dumped [here](https://github.com/ruby-git/ruby-git/actions/runs/9248412388/job/25438675171?pr=724)  with that I dumped [here](https://github.com/ruby-git/ruby-git/actions/runs/9248489248/job/25438861693?pr=724), you'll see that it's not the _gem_ that's producing mistaken results, _git itself_ seems to have a misapprehension of the state of the system, when ignoreCase is set. Which.. obviously makes me think I'm just failing to grasp something, but I can't think what it might be - these are ubuntu containers, so the the FS itself is clearly case-sensitive, and I updated to git 2.45 locally for comparison's sake.

But in the first of those links, after I move `scott/text.txt` to `scott/TEXT.txt`, it (the git cli) displays the former as _deleted_, not just renamed (or better, unchanged). And in the latter link, when I start moving it to `scott/TEXT2.txt`, I can see it again. Is `ignoreCase` intended only to be used in conjunction with case-insensitive file systems, perhaps?

Ah, that is in fact the case! https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreignoreCase

I find it odd that the test passed locally, since I am on a case-sensitive file-system here as well, but I'm guessing the OSX version of git has some differences in this area, since macs were CI by default until a few years ago.

At any rate, that makes writing tests that actually _exercise_ this behavior.. problematic :-\

jcouball (approved):
Thanks again, Eric.

It seems awfully inefficient to downcase all keys in the collection every time you call one of these predicate methods. But with the tests we can optimize later.

Looks good 👍🏼

nevinera:
> Thanks again, Eric.
> 
> It seems awfully inefficient to downcase all keys in the collection every time you call one of these predicate methods. But with the tests we can optimize later.
> 
> Looks good 👍🏼 

I actually initially had it memoized, and had started down the path of memoizing all the other methods, but it was making the PR quite large. The existing predicates actually are far worse already, as they generate the corresponding collection dynamically in each call by iterating the `@files`, then check membership.

I decided that meant that the working set of items must typically be small enough not to worry about it, but I'd be happy to roll up a PR memoizing all the things, if that's of interest?

nevinera:
@jcouball - here are the performance improvements, when you have a minute: https://github.com/ruby-git/ruby-git/pull/729

jcouball:
I am interested but can live without it for now. Let's see if anyone complains.

PR #728, opened 2024-06-01
Supply all of the _specific_ color options too

nevinera:
Previously, we were supplying `color.ui=false`, but if the local gitconfig specified any of the more specific options (like `color.diff`), those would take precedence. This updates our command-runner to always supply all of the specific color options as false as well, so that we definitely get a color-free output suitable for parsing.

Updating the fixture causes several of the existing specs to fail without the code changes (especially the `diff` tests), which is probably an appropriate level of testing - I doubt all of these config options are actually necessary, but I don't see any harm in including them.

This should replace #723, since that won't be necessary anymore. We don't currently have any tests checking the actual contents of the commands being run, but I _could_ rig up a test that supplies a `.git/config`.


jcouball (approved):
Ok, this seems to do the trick. Looks great!

PR #729, opened 2024-06-01
Memoize all of the significant calls in Git::Status

nevinera:
When the status has many entries, there were substantial inefficiencies in this class - calling predicates like `changed?(filename)` would iterate the status, constructing a transient `changed` subhash, then test that subhash to see if the file in question was in it (for example).

After this change, it will _keep_ those sub-hashes for reuse on the Status instance, as well as `downcase`d versions if they happen to get requested (by case-insensitive calls).

### DIFF
--- a/lib/git/status.rb
+++ b/lib/git/status.rb
@@ -22,7 +22,7 @@ module Git
     #
     # @return [Enumerable]
     def changed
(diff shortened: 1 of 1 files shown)


jcouball (approved):
Yes. I can see what you said about the complexity. But this is great! Thank you again.

PR #737, opened 2024-08-25
Refactor Git::Lib#rev_parse(revision) and sanitize its input

jcouball:
Make the following changes to Git::Lib#rev_parse:

* Remove optimizations that try to calculate the resulting SHA instead of calling `git rev-parse`. This library should not reimplement git functionality if possible. Git could change and then the assumptions in these optimizations would no longer be valid (e.g. SHAs are 40 digit hex numbers).

  There is a chance that this could break existing uses if the optimization implementation is already different than the implementation of `git rev-parse`

* Validate that the reference passed in does not look like a command-line option (i.e. starts with a hyphen).

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -311,16 +311,31 @@ module Git
       process_commit_log_data(full_log)
     end
 
-    def revparse(string)
-      return string if string =~ /^[A-Fa-f0-9]{40}$/  # passing in a sha - just no-op it
-      rev = ['head', 'remotes', 'tags'].map do |d|
-        File.join(@git_dir, 'refs', d, string)
-      end.find do |path|
-        File.file?(path)
-      end
-      return File.read(rev).chomp if rev
-      command('rev-parse', string)
+    # Verify and resolve a Git revision to its full SHA
+    #
+    # @see https://git-scm.com/docs/git-rev-parse git-rev-parse
+    # @see https://git-scm.com/docs/git-rev-parse#_specifying_revisions Valid ways to specify revisions
+    # @see https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt-emltrefnamegtemegemmasterememheadsmasterememrefsheadsmasterem Ref disambiguation rules
+    #
+    # @example
+    #   lib.rev_parse('HEAD') # => '9b9b31e704c0b85ffdd8d2af2ded85170a5af87d'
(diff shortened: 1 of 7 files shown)


lib/git/lib.rb
```diff
@@ -311,17 +311,32 @@ def full_log_commits(opts = {})
...
+    #
+    # @return [String] the full commit hash
+    #
+    # @raise [Git::FailedError] if the revision cannot be resolved
+    # @raise [ArgumentError] if the revision is a string starting with a hyphen
+    #
+    def rev_parse(revision)
+      assert_args_are_not_options('rev', revision)
+
+      command('rev-parse', revision)
```
mpapis, line 334: Hey @jcouball, thank you for this project.

I wonder if this command could be adjusted like this:
```suggestion
      command('rev-parse', '--revs-only', '--end-of-options', revision, '--')
```
This changes the `exception.result.stderr` to:
```bash
$ git rev-parse --revs-only --end-of-options v3 --
fatal: bad revision 'v3'
```
from:
```bash
$ git rev-parse v3
v3
fatal: ambiguous argument 'v3': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
Which left me confused and investigating first time I [encountered missing revision name](https://github.com/mpapis/uncov/commit/23c312c29a7594ef8c05ef269b395d398e4c6cf7#diff-9e2a28deb4619667b946a3a8dd073c3fa86839ff5f829b27d339a85bc4167ebbR47-R50).

If you are up for this change I can prepare a PR for it, WDYT?

jcouball: Yes, a PR would be appreciated!

I do have a couple of caveats. I’m not sure if they can be fully addressed, but I’m happy to take the PR either way.

Appending `--` at the end of the command seems a bit ambiguous when `--revs-only` is also used. From what I’ve observed, Git appears to ignore anything after `--` in this context — even if you provide a path (existing or not), the output remains the same.

That said, the presence of `--` does change the error messaging. As you pointed out, it produces a more helpful error when the input is invalid.

This raises a couple of questions:

* Can we rely on this behavior remaining consistent long-term?
* Is there a risk that combining --revs-only with -- will be considered an error in future Git versions?

mpapis: https://github.com/ruby-git/ruby-git/pull/789 - I think that `--` behavior is so deep at core of git that it will never change.

PR #744, opened 2024-09-01
Sanatize object ref sent to cat-file command

jcouball:
Sanitize the object reference sent to the cat-file command.

Make sure the object reference sent to cat-file commands do not start with a hyphen so they are not mistaken for command line options.

Also normalize the cat-file commands in accordance with [this project's design philosophy](https://github.com/ruby-git/ruby-git/blob/master/CONTRIBUTING.md#design-philosophy). This includes adding the following commands (and aliases):

* #cat_file_contents was aliased to #object_contents
* #cat_file_type was aliased to #object_type
* #cat_file_size was aliased to #object_size
* #cat_file_commit was aliased to #commit_data
* #cat_file_tag was aliased to #tag_data

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -650,7 +650,7 @@ module Git
     end
 
     def cat_file(objectish)
-      self.lib.object_contents(objectish)
+      self.lib.cat_file(objectish)
     end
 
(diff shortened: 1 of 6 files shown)


jcouball:
Will fix windows build in a separate PR.

PR #758, opened 2024-12-04
fix: open3 errors on binary paths with spaces

### LINKED ISSUE #757: Open3.capture2 causes failures in open and binary_version
### Subject of the issue
The use of `Open3.capture2` results in failures when binary paths with a space are used due to how `Open3` parses commands.

### Your environment
- ruby: 3.3.5
- ruby-git: 2.3.2
- git: 2.44.0

We are running ruby through cygwin but calling a native windows git binary which lives in "Program Files" which is how we caught this issue.

### Steps to reproduce
Put the git binary into a path with a space.
Configure the git binary
```
::Git.configure do |config|
  config.binary_path = binary_path
end
```

Attempt to check the binary version or open an existing git project.
```
::Git.binary_version
::Git.open("some existing git folder")
```

The binary version command will fail

### Expected behaviour
Commands to succeed and return the correct version or open the existing git project.

### Actual behaviour
Commands failed. Example of the binary version failure:
```
Failed to get git version: pid 752 exit 127
sh: /cygdrive/c/Program: No such file or directory 
```

bcg00ding:
# Description
Resolves #757 which is caused by `Open3` splitting the string by spaces. This is an issue when Git binaries exist in a path containing spaces like `C:\Program Files\Git\git.exe`

By splitting to `args` instead of a single string `Open3` is able to properly execute the command against the git binary. Switching to `capture2e` is required because `capture2` generates an error when `2&>1` is used. `capture2e` redirects `stderr` to `stdout` for us.

Because there is a possibility that `Open3` will now fail because the binary doesn't exist we have to handle the execution of `Open3` in a rescue block and raise the correct error type.

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -37,9 +37,15 @@ module Git
     end
 
     def self.binary_version(binary_path)
-      git_cmd = "#{binary_path} -c core.quotePath=true -c color.ui=false version 2>&1"
-      result, status = Open3.capture2(git_cmd)
-      result = result.chomp
+      result = nil
+      status = nil
+
+      begin
+        result, status = Open3.capture2e(binary_path, "-c", "core.quotePath=true", "-c", "color.ui=false", "version")
+        result = result.chomp
+      rescue Errno::ENOENT
+        raise RuntimeError, "Failed to get git version: #{binary_path} not found"
+      end
 
       if status.success?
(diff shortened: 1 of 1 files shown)


jcouball (approved):
Thank you for the contribution and thank you for fixing this bug.

Your changes look good. I have added tests which I will add in the next PR.

PR #789, opened 2025-05-14
Improved error message of rev_parse

mpapis:
# Description

Improved error message of rev_parse

As described by git-rev-parse:
```
  Many Git porcelainish commands take mixture of flags (i.e. parameters
  that begin with a dash -) and parameters meant for the underlying git
  rev-list command they use internally and flags and parameters for the
  other commands they use downstream of git rev-list. This command is
  used to distinguish between them.
```

Using the `--` to separate revisions from paths is at the core of git. I do not think this behavior will ever change.

The message without the extra parameters:
```
  fatal: ambiguous argument 'v3': unknown revision or path not in the
  working tree.
  Use '--' to separate paths from revisions, like this:
  'git <command> [<revision>...] -- [<file>...]'
```

The message with new parameters:
```
  fatal: bad revision 'NOTFOUND'
```

I think it's way more descriptive.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -333,7 +333,7 @@ module Git
     def rev_parse(revision)
       assert_args_are_not_options('rev', revision)
 
-      command('rev-parse', revision)
+      command('rev-parse', '--revs-only', '--end-of-options', revision, '--')
     end
 
     # For backwards compatibility with the old method name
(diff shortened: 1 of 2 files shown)


jcouball (approved):
Looks good... short and sweet!

jcouball:
@mpapis I just pushed 3.0.1 to RubyGems with this fix.

PR #810, opened 2025-07-02
Upgrade to ProcessExecuter 4.x

jcouball:
Update the dependency to ProcessExecuter 4.x and modify the code to accommodate changes in the API, ensuring backward compatibility where necessary.


jcouball:
My IDE is still insisting that master is the default branch.

PR #891, opened 2026-01-15
ci: add experimental TruffleRuby container job for performance comparison

jcouball:
## Summary

This PR adds an experimental parallel CI job to test whether using a pre-built TruffleRuby container is faster than using `setup-ruby` to download and install TruffleRuby.

## Motivation

TruffleRuby downloads are large (~250MB+) and can take 2-5 minutes even with caching. Using a pre-built container could significantly reduce CI time.

## What's Added

A new `build-truffleruby-container` job that:
- **Runs in parallel** with existing matrix jobs (won't block CI)
- **Uses container**: `ghcr.io/graalvm/truffleruby-community:24.2.1`
- **Skips setup-ruby**: TruffleRuby is pre-installed in the container
- **Marked experimental**: `continue-on-error: true` so failures won't block CI

## Side-by-Side Comparison

After this PR runs, we can compare:

| Job | Approach | Expected Setup Time |
|-----|----------|-------------------|
| `Ruby truffleruby-24.2.1 on ubuntu-latest` | setup-ruby (existing) | 2-5 minutes |
| `Ruby truffleruby-24.2.1 (container) on ubuntu-latest` | container (new) | <1 minute |

## Next Steps

If the container job proves:
- ✅ **Faster**: Replace the matrix entry with this approach
- ✅ **More reliable**: Less download failures
- ❌ **Problematic**: Close this PR and keep using setup-ruby

## Testing

This PR will trigger both jobs so we can see the performance difference immediately in the Actions tab.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -53,3 +53,37 @@ jobs:
 
       - name: Test Gem
         run: bundle exec rake test:gem
+
+  # Experimental: TruffleRuby with pre-built container (for speed comparison)
+  build-truffleruby-container:
+    name: Ruby truffleruby-24.2.2 (container) on ubuntu-latest
+
+    # Skip this job if triggered by a release PR
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && !startsWith(github.event.pull_request.head.ref, 'release-please--'))
+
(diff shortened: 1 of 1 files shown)


jcouball:
The TruffleRuby build in the container actually took about 50s longer:

with setup-ruby action: 2m 54s
with TruffleRuby container: 3m 42s

So not faster. Will close this without merging.

PR #1159, opened 2026-03-30
fix: return empty result from git log on repo with no commits

### LINKED ISSUE #1155: Handle empty list of commits in new repository
I started to create a PR for this issue as it is simple and then I saw in `CONTRIBUTING.md`

> ... it does not introduce unnecessary abstraction layers or modify Git's core functionality.

so I think that raising an issue to ask is more sensible.

My first attempt with this gem was to try

```ruby
repo = Git.open(File.expand_path('source', __dir__))
repo.log.execute
```

on a new and completely empty repository. This results in an exception being thrown as `git log` exits with a fatal error.

```bash
$ git log
fatal: your current branch 'master' does not have any commits yet
```

 I would expect this to fail silently and return an empty array but I appreciate that this differs from the functionality of `git`. Is this a change that you would consider?

jcouball:
## Summary

When `Git::Log#execute` is called on a newly initialized repository that has no commits, `git log` exits with status 128 and the message "does not have any commits yet". This change rescues that specific error in `log_commits` and `full_log_commits`, returning an empty array instead of raising `Git::FailedError`.

## Changes

- Add `log_or_empty_on_unborn` helper in `Git::Lib` that rescues `Git::FailedError` when exit status is 128 and stderr matches "does not have any commits yet"
- Wrap `log_commits` and `full_log_commits` with this helper to return `[]` on empty repos
- Add unit tests for both `Git::Log#execute` and `Git::Lib#log_commits`/`#full_log_commits` on empty repos

Fixes #1155

### DIFF
--- a/bin/docker-git
+++ b/bin/docker-git
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -e
+
+usage() {
+  echo "Build and run a Docker container with a specific version of git."
+  echo ""
+  echo "Usage: $(basename "$0") <version>"
+  echo ""
+  echo "Example: $(basename "$0") 2.28.0"
+  exit 1
+}
+
+[ -z "$1" ] && usage
+
+GIT_VERSION="$1"
+SCRIPT_DIR=$(dirname "$0")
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
Thank you. I pulled this branch to test and it behaves in exactly the way I am hoping for in #1155.

PR #1204, opened 2026-04-11
docs: remove overly strict flag-form rules from review-arguments-dsl checklist

jcouball:
## Summary

Removes two checklist rules from the `review-arguments-dsl` skill that were too prescriptive and caused unnecessary friction during command reviews:

- **Negatable options must document both emitted forms** — the `false`/`--no-foo` form doesn't need to be explicitly documented in every `@option`
- **`@option` descriptions must use the emitted long flag form** — short flag aliases are acceptable in prose when they are the more commonly referenced form

## Changes

- `.github/skills/review-arguments-dsl/CHECKLIST.md`: remove the two overly strict sections
- `lib/git/commands/branch/list.rb`: update `:verbose` YARD docs to use `-v` (the common alias), consistent with the relaxed rule


jcouball:
Decided to leave these rules in.

PR #1296, opened 2026-05-12
fix(remote_operations): raise ArgumentError when :ref is supplied without an explicit remote (#1291)

### LINKED ISSUE #1291: bug: `fetch(ref: ...)` without a remote silently passes refspec as repository operand
## Summary

When `fetch` is called in the hash-only form (no explicit remote) with a `:ref` option, the refspec ends up as the first positional argument to `git fetch`, which git interprets as the `<repository>` operand rather than a refspec. This causes a confusing failure or, in rare cases, unintended behavior.

## Reproduction

```ruby
repo.fetch(ref: 'refs/heads/feature')
```

This generates:

```shell
git fetch refs/heads/feature
```

instead of the presumably intended (which has no valid git equivalent without a remote):

```shell
git fetch <remote> refs/heads/feature
```

Git then reports:

```
fatal: 'refs/heads/feature' does not appear to be a git repository
```

## Root cause

In both `Git::Lib#fetch` (on `main`) and `Git::Repository::RemoteOperations#fetch` (introduced in #1290), the hash-dispatch pattern sets `remote = nil` when only an options hash is passed:

```ruby
def fetch(remote = 'origin', opts = {})
  if remote.is_a?(Hash)
    opts = remote
    remote = nil          # remote is now nil
  end
  # ...
  refspecs = Array(opts.delete(:ref)).compact
  positionals = [*([remote] if remote), *refspecs]
  # When remote is nil and ref is set:
  # positionals = ['refs/heads/feature']  ← refspec promoted to repository slot
```

The `OperandAllocator` in `Git::Commands::Fetch` then binds the refspec string to the `:repository` operand, not `:refspec`.

## Fix

Add a guard in the facade method raising `ArgumentError` when `:ref` is provided without an explicit remote:

```ruby
if remote.nil? && opts.key?(:ref)
  raise ArgumentError, ':ref requires an explicit remote; use fetch(remote, ref: ...) form'
end
```

## Existing test to update

`spec/unit/git/repository/remote_operations_spec.rb` contains a test that documents the current (buggy) behaviour with a reference to this issue:

```ruby
# Known bug: when :ref is supplied without an explicit remote, the refspec is
# promoted to the :repository operand slot, causing git to treat it as a remote
# name/URL. This will be fixed in issue #1291.
context 'with :ref and no remote (Hash form)' do
  it 'passes the refspec as the repository positional (pre-#1291 behaviour)' do
    ...
  end
end
```

This example must be replaced with one that asserts `ArgumentError` is raised.

## Notes

- This bug predates the facade extraction in #1290 — it originates in `Git::Lib#fetch` on `main`
- The fix should be applied in the facade layer, not the command layer (`Git::Commands::Fetch` cannot distinguish a legitimate repository URL from a misrouted refspec)
- The fix should also be applied to `Git::Lib#fetch` on `main` while it remains in place (until Phase 4 of the architectural redesign deletes `Git::Lib`)

SAY-5:
Closes #1291.

`Git::Repository::RemoteOperations#fetch` shifts a single hash argument into `opts` and sets `remote = nil`. The subsequent `positionals = [*([remote] if remote), *refspecs]` then leaves the refspec as the first positional, which the command layer binds to the `:repository` operand. `git fetch refs/heads/main` is then interpreted as a remote name and fails with "does not appear to be a git repository".

Adds the guard suggested in the issue: if `remote` is nil after the hash shift and `:ref` is present, raise `ArgumentError` with a hint pointing at the two-arg form.

The existing pre-fix unit test in `spec/unit/git/repository/remote_operations_spec.rb` documented the buggy behaviour and explicitly tagged itself for removal here; it now asserts the `ArgumentError` instead and verifies the command layer is never reached.

### DIFF
--- a/lib/git/repository/remote_operations.rb
+++ b/lib/git/repository/remote_operations.rb
@@ -110,19 +110,18 @@ module Git
       #     from each branch tip (`--depth=N`)
       #
       #   @option opts [String, Array<String>] :ref (nil) one or more refspecs to
-      #     fetch; forwarded as positional arguments after the remote name
+      #     fetch; forwarded as positional arguments after the remote name. An
+      #     explicit `remote` is required when `:ref` is given.
       #
       #   @return [String] the merged stdout from the fetch command
       #
-      #   @raise [ArgumentError] when unsupported option keys are provided
+      #   @raise [ArgumentError] when unsupported option keys are provided or `:ref`
+      #     is supplied without an explicit remote
       #
       #   @raise [Git::FailedError] when git exits with a non-zero status
       #
       def fetch(remote = 'origin', opts = {})
-        if remote.is_a?(Hash)
-          opts = remote
-          remote = nil
-        end
+        remote, opts = Private.resolve_fetch_target(remote, opts)
 
         opts = Private.normalize_fetch_keys(opts)
         SharedPrivate.assert_valid_opts!(FETCH_ALLOWED_OPTS, **opts)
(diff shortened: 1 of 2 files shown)


lib/git/repository/remote_operations.rb
jcouball: Update to mention the new ArgumentError cause. Something like "when unsupported option keys are provided or `:ref` is supplied without an explicit remote"

jcouball: Should mention that a `remote` is required when a `:ref` is given.

jcouball (commented):
Requesting two small corrections. With these corrections, I can merge the change.

jcouball (approved):
Thanks for the contribution @SAY-5, this looks great.

PR #1331, opened 2026-05-20
fix: preserve colon-number filenames in grep results

### LINKED ISSUE #1324: bug: grep parser misparses filenames containing `:<digits>:` patterns
## Summary

`Git::Parsers::Grep.parse` (and its predecessor `Git::Lib#parse_grep_output` in 4.x) uses a non-greedy regex to split `git grep --line-number` output:

```ruby
match = line.match(/\A(.*?):(\d+):(.*)/)
```

Because `(.*?)` stops at the *first* `:(\d+):` occurrence, any filename that contains a colon-digits-colon sequence is misparsed.

## Example

Given a filename `foo:42:bar.rb`, the raw `git grep --line-number` output line looks like:

```
foo:42:bar.rb:5:matched text
```

The regex captures:
| Group | Expected | Actual |
|-------|----------|--------|
| filename | `foo:42:bar.rb` | `foo` |
| line_number | `5` | `42` |
| text | `matched text` | `bar.rb:5:matched text` |

## Origin

This bug was present in `Git::Lib#parse_grep_output` (4.x) and was faithfully preserved during the parser extraction in PR #1323. It was first flagged as a code-review comment on that PR.

Git explicitly permits colons in filenames, so this is a correctness gap.

## Possible fixes

1. **Use NUL-delimited output (`-z`)** — pass `-z` to `git grep` so fields are separated by NUL bytes instead of colons. This is the most robust fix and is independent of filename content.
2. **Document the known limitation** — add a note to `Git::Parsers::Grep` and the public `grep` method that filenames containing `:<digits>:` are not supported.

Option 1 is preferred but may be a minor breaking change if callers depend on the exact hash key format returned by `grep`.

## References

- PR #1323 — parser extraction (where the issue was first flagged)
- `lib/git/parsers/grep.rb`
- `Git::Repository::ObjectOperations#grep`

puneetdixit200:
## Summary
- run public grep calls with `--null` so git separates path and line fields unambiguously
- add a NUL-delimited grep parser for `treeish:filename\0line\0text` output
- cover filenames containing `:<digits>:` at parser and repository integration levels

Closes #1324.

## Verification
- `PATH="/opt/homebrew/opt/ruby/bin:$PATH" bundle exec rspec spec/unit/git/parsers/grep_spec.rb spec/integration/git/repository/object_operations_spec.rb --format progress` -> 100 examples, 0 failures
- `PATH="/opt/homebrew/opt/ruby/bin:$PATH" bundle exec rubocop lib/git/parsers/grep.rb lib/git/repository/object_operations.rb lib/git/lib.rb spec/unit/git/parsers/grep_spec.rb spec/integration/git/repository/object_operations_spec.rb` -> 5 files inspected, no offenses detected

## AI-assisted work
This patch was prepared with AI assistance and manually reviewed. I verified the failure first with a spec for the colon-number filename case, then reran the targeted specs and style checks locally.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -7,6 +7,7 @@ require 'git/command_line'
 require 'git/errors'
 require 'git/parsers/branch'
 require 'git/parsers/fsck'
+require 'git/parsers/grep'
 require 'git/parsers/stash'
 require 'git/parsers/tag'
 require 'git/url'
@@ -801,7 +802,9 @@ module Git
 
       opts = normalize_grep_opts(opts)
       object = opts.delete(:object) || 'HEAD'
-      result = Git::Commands::Grep.new(self).call(object, pattern:, **opts, no_color: true, line_number: true)
+      result = Git::Commands::Grep.new(self).call(
+        object, pattern:, **opts, no_color: true, line_number: true, null: true
+      )
       exitstatus = result.status.exitstatus
 
       # Exit status 1 with empty stderr means no lines matched (not an error)
@@ -810,7 +813,7 @@ module Git
       # Exit status 1 with non-empty stderr is a real error (e.g. bad object reference)
       raise Git::FailedError, result if exitstatus == 1
 
-      parse_grep_output(result.stdout.split("\n"))
+      parse_grep_output(result.stdout)
     end
 
     # Validate that the given arguments cannot be mistaken for a command-line option
@@ -2505,14 +2508,8 @@ module Git
       opts
     end
 
-    def parse_grep_output(lines)
-      lines.each_with_object(Hash.new { |h, k| h[k] = [] }) do |line, hsh|
-        match = line.match(/\A(.*?):(\d+):(.*)/)
-        next unless match
-
-        _full, filename, line_num, text = match.to_a
-        hsh[filename] << [line_num.to_i, text]
-      end
+    def parse_grep_output(output)
+      Git::Parsers::Grep.parse(output)
     end
 
     def parse_diff_stats_output(lines)
--- a/lib/git/parsers/grep.rb
+++ b/lib/git/parsers/grep.rb
@@ -4,9 +4,8 @@ module Git
   module Parsers
     # Parser for `git grep` output
     #
-    # Provides a class method that transforms raw `git grep` output lines into
-    # a structured Hash consumed by the `Git::Repository::ObjectOperations`
-    # facade.
+    # Provides a class method that transforms raw `git grep --null` output into a
+    # structured Hash consumed by the `Git::Repository::ObjectOperations` facade.
     #
     # This parser is a pure text transformer with no exit-status logic. The
     # calling facade is responsible for interpreting the command's exit status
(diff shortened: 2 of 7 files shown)


lib/git/parsers/grep.rb
```diff
@@ -38,6 +38,28 @@ def parse(lines)
           hsh[filename] << [line_num.to_i, text]
         end
       end
```
jcouball: This PR leaves `.parse` as dead code. Please remove this method and rename `.parse_null_delimited` to `.parse`. Adjust tests accordingly. I don't think you'll need to delete existing tests... just make sure they work.

lib/git/parsers/grep.rb
```diff
@@ -38,6 +38,28 @@ def parse(lines)
...
+      # contain `:<digits>:` from being confused with the line-number delimiter.
+      #
+      # @param output [String] raw output from `git grep --null --line-number`
+      #
+      # @return [Hash<String, Array<Array(Integer, String)>>] hash mapping
+      #   `"treeish:filename"` keys to arrays of `[line_number, text]` pairs
+      #
+      # @api private
+      #
+      def parse_null_delimited(output)
```
jcouball: Rename to `parse`

lib/git/repository/object_operations.rb
```diff
@@ -583,7 +583,7 @@ def parse_grep_result(result)
           return {} if exitstatus == 1 && result.stderr.empty?
           raise Git::FailedError, result if exitstatus == 1
 
-          Git::Parsers::Grep.parse(result.stdout.split("\n"))
+          Git::Parsers::Grep.parse_null_delimited(result.stdout)
```
jcouball: This should be renamed to `parse`

lib/git/lib.rb
```diff
@@ -2505,14 +2508,8 @@ def normalize_grep_opts(opts)
...
-    def parse_grep_output(lines)
-      lines.each_with_object(Hash.new { |h, k| h[k] = [] }) do |line, hsh|
-        match = line.match(/\A(.*?):(\d+):(.*)/)
-        next unless match
-
-        _full, filename, line_num, text = match.to_a
-        hsh[filename] << [line_num.to_i, text]
-      end
+    def parse_grep_output(output)
+      Git::Parsers::Grep.parse_null_delimited(output)
```
jcouball: Use #parse instead.

spec/unit/git/parsers/grep_spec.rb
```diff
@@ -33,6 +33,17 @@
       end
     end
 
+    context 'with NUL-delimited grep output' do
+      subject(:result) do
+        nul = "\0"
+        described_class.parse_null_delimited("HEAD:src/foo:42:bar.rb#{nul}5#{nul}matched text\n")
```
jcouball: Two things:

1. It is strange that you nest a test for `#parse_null_delimited` in the `describe '#parse '` block. If you rename `#parse_null_delimited` to `#parse`, then this is a moot point.

2. Please adjust test nesting as follows:

```ruby
describe '#parse' do
  ...
  context 'with a matching filename containing a colon-number-colon pattern' do
    ... setup the test data ...
    it 'preserves the filename' do
      ... expectation ...
    end
  end
end
```

jcouball (changes requested):
I found the following issues:

- Old `.parse` method is dead code: delete the existing `.parse` and rename `.parse_nil_delimited` to `.parse`. Keep existing tests (adjustments in the tests may be needed).
- Adjust the nesting of the new test as described in code comment.

lib/git/parsers/grep.rb
puneetdixit200: Implemented in `fd0ab7e`: removed the old parser, renamed the null-delimited parser to `.parse`, and adjusted the tests.

lib/git/parsers/grep.rb
puneetdixit200: Renamed to `.parse` in `fd0ab7e`.

lib/git/repository/object_operations.rb
puneetdixit200: Updated this call site to use `.parse` in `fd0ab7e`.

lib/git/lib.rb
puneetdixit200: Updated this to use `.parse` in `fd0ab7e`.

spec/unit/git/parsers/grep_spec.rb
puneetdixit200: Adjusted the test nesting under `describe "#parse"` in `fd0ab7e`.

jcouball (approved):
Much better.

PR #1335, opened 2026-05-20
fix: parse grep output with null delimiters

### LINKED ISSUE #1324: bug: grep parser misparses filenames containing `:<digits>:` patterns

puneetdixit200:
# Description

Fixes #1324 by making repository grep calls request NUL-delimited output from git and by parsing the file\0line\0text format. This keeps filenames such as `src/foo:42:bar.rb` from being split at the wrong colon sequence, while preserving matched text containing `:13:`.

The legacy `Git::Lib#grep` parser now delegates to `Git::Parsers::Grep` so both grep paths share the same output parsing.

AI assistance disclosure: this PR was assisted by OpenAI GPT-5. I reviewed the generated changes in this checkout and verified them with the checks below.

## Checklist

- [x] I reviewed and applied the project AI Policy. The AI-assisted changes were inspected and validated for quality, security, and licensing concerns.
- [ ] The human submitter has run `bundle exec rake` locally on this branch before moving this PR out of draft.
- [x] Documentation updated where applicable. No README/YARD update was needed because this fixes internal parsing behavior without changing the public API.

## Validation

- `bundle exec rspec spec/unit/git/parsers/grep_spec.rb spec/unit/git/repository/object_operations_spec.rb spec/unit/git/lib_command_spec.rb spec/integration/git/repository/object_operations_spec.rb` passed: 300 examples, 0 failures.
- `bundle exec rubocop lib/git/lib.rb lib/git/parsers/grep.rb lib/git/repository/object_operations.rb spec/unit/git/parsers/grep_spec.rb spec/unit/git/repository/object_operations_spec.rb spec/unit/git/lib_command_spec.rb spec/integration/git/repository/object_operations_spec.rb` passed: 7 files inspected, no offenses.
- `bundle exec rake` passed in a Ruby 3.4 Docker container: Test::Unit, RSpec unit/integration, RuboCop, YARD, YARD examples, and gem build.
- `npx commitlint --from HEAD~1 --to HEAD` passed.

### DIFF
--- a/lib/git/lib.rb
+++ b/lib/git/lib.rb
@@ -7,6 +7,7 @@ require 'git/command_line'
 require 'git/errors'
 require 'git/parsers/branch'
 require 'git/parsers/fsck'
+require 'git/parsers/grep'
 require 'git/parsers/stash'
 require 'git/parsers/tag'
 require 'git/url'
@@ -801,7 +802,9 @@ module Git
 
       opts = normalize_grep_opts(opts)
       object = opts.delete(:object) || 'HEAD'
-      result = Git::Commands::Grep.new(self).call(object, pattern:, **opts, no_color: true, line_number: true)
+      result = Git::Commands::Grep.new(self).call(
+        object, pattern:, **opts, no_color: true, line_number: true, null: true
+      )
       exitstatus = result.status.exitstatus
 
       # Exit status 1 with empty stderr means no lines matched (not an error)
@@ -2506,13 +2509,7 @@ module Git
     end
 
     def parse_grep_output(lines)
-      lines.each_with_object(Hash.new { |h, k| h[k] = [] }) do |line, hsh|
-        match = line.match(/\A(.*?):(\d+):(.*)/)
-        next unless match
-
-        _full, filename, line_num, text = match.to_a
-        hsh[filename] << [line_num.to_i, text]
-      end
+      Git::Parsers::Grep.parse(lines)
     end
 
     def parse_diff_stats_output(lines)
--- a/lib/git/parsers/grep.rb
+++ b/lib/git/parsers/grep.rb
@@ -31,12 +31,25 @@ module Git
       #
       def parse(lines)
         lines.each_with_object(Hash.new { |h, k| h[k] = [] }) do |line, hsh|
-          match = line.match(/\A(.*?):(\d+):(.*)/)
-          next unless match
+          filename, line_num, text = parse_line(line)
+          next unless filename && line_num && text
 
-          _full, filename, line_num, text = match.to_a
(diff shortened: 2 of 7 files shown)


jcouball:
Thank you for tackling #1324. The core strategy (--null output, NUL-split parsing) is the right approach. However there are a few issues that need to be addressed, and I want to flag that PR #1331 (also by you) covers the same fix. Please close one of the two PRs and consolidate your changes.

In my assessment, #1331 was the better of the two PRs and #1335 regresses in some ways. I feel like you are leaning too heavily on AI responses and not understanding the generated code and what the problems are with it. 

IMHO, you should close #1335 and continue with #1331 following the advice I gave there. Additionally, you will need to rebase to origin/main and resolve the conflicts you find (make sure not to delete code landed in previous PRs).

In this PR:

- `parse_colon_delimited_line` can never be reached. Both call sites now pass `null: true` unconditionally, so git will always produce NUL-delimited output (which is what is wanted). The colon-delimited fallback inside `parse` is now dead code. It should be removed.
- The Windows CI build failed because you didn't skip the test for filenames with colons like you did in #1331.

PR #1422, opened 2026-06-12
feat: add deprecated config_get/list/set aliases to Git::Repository::Configuring

jcouball:
# Description

Add deprecated forwarding wrappers `config_get`, `config_list`, and `config_set`
to `Git::Repository::Configuring`, and the corresponding `Git::Base` delegators.

## Context

`Git::Repository::Configuring#config()` already covers all three cases via its
three-way dispatch:

- `config` → list all entries (`Hash`)
- `config(name)` → get one entry (`String`)
- `config(name, value)` → set one entry

Callers who used `g.lib.config_get/list/set` directly need a supported migration
path: they can drop the `.lib.` prefix and get a runtime deprecation warning
rather than a hard `NoMethodError`. All three aliases are removed in v6 after
a full deprecation cycle.

This is PR 5h-2 from `redesign/c1c2_bucket6_lib_orphans.md` §3.2.

### `lib/git/repository/configuring.rb`

Three deprecated methods added after `global_config_set`, before `Private`:

- `config_get(name)` — emits `Git::Deprecation.warn`, delegates to `config(name)`
- `config_list` — emits `Git::Deprecation.warn`, delegates to `config`
- `config_set(name, value, opts = {})` — emits `Git::Deprecation.warn`, delegates to `config(name, value, opts)`

Each method has full YARD documentation including `@deprecated`, `@param`,
`@return`, `@raise`, and `@see #config` tags.

### `lib/git/base.rb`

Three delegators added after `global_config_set`:

- `config_get(name)` → `facade_repository.config_get(name)`
- `config_list` → `facade_repository.config_list`
- `config_set(name, value, opts = {})` → `facade_repository.config_set(name, value, opts)`

### `redesign/c1c2_audit.md`

- §7.3: three `config_get/list/set` rows updated from `🔍 human decision` to `✅ promoted as deprecated methods`
- §7.5: 🔍 human decision count decreased from 11 to 8

### `redesign/c1c2_bucket6_lib_orphans.md`

- §3.2: status line added — ✅ Implemented

### `UPGRADING.md`

No change needed — both the `g.lib.config_*` rows (§ "Methods with a direct
replacement") and the `g.config_*` rows (§ "Deprecated methods") were already
present from a prior commit.

## Test coverage

- **`spec/unit/git/repository/configuring_spec.rb`** — 6 new examples (2 per method):
  one asserting `Git::Deprecation.warn` is called with the method name, one
  asserting correct delegation and return value.

- **`spec/unit/git/base_spec.rb`** — 3 new examples (1 per method), following
  the `global_config_get/list/set` delegator test pattern.

No integration tests added — the deprecated wrappers delegate directly to
`config()`, which is already integration-tested.

## Checklist

### DIFF
--- a/lib/git/base.rb
+++ b/lib/git/base.rb
@@ -247,6 +247,18 @@ module Git
       facade_repository.global_config_set(name, value)
     end
 
+    def config_get(name)
+      facade_repository.config_get(name)
+    end
+
+    def config_list
+      facade_repository.config_list
+    end
+
+    def config_set(name, value, opts = {})
+      facade_repository.config_set(name, value, opts)
+    end
+
     # Returns a reference to the working directory
     #
     # @example
--- a/lib/git/repository/configuring.rb
+++ b/lib/git/repository/configuring.rb
@@ -60,7 +60,10 @@ module Git
       #   @param name [String] the dotted config key to write (e.g.
       #     `"user.name"`)
       #
-      #   @param value [String] the value to assign
+      #   @param value [#to_s] the value to assign; must not be `nil` (a `nil`
+      #     value is treated as "no value" and routes to the get overload).
+      #     Any non-nil object is converted to a String via `#to_s` before
+      #     being passed to git
       #
       #   @param options [Hash] options for the set operation
       #
@@ -75,7 +78,7 @@ module Git
       # @raise [Git::FailedError] if git exits with a non-zero exit status
       #
       def config(name = nil, value = nil, options = {})
-        if name && value
+        if !name.nil? && !value.nil?
           Private.config_set(@execution_context, name, value, **options)
         elsif name
           Private.config_get(@execution_context, name)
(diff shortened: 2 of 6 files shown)


jcouball:
Round 6 fix: changed the `config_set` deprecation warning from `config(name, value, opts)` to `config(name, value, options)` to match the public `#config` signature (`options = {}`) rather than exposing the local variable name `opts`. Tightened the unit test to assert the warning includes `config(name, value, options)` so this can't regress. Rake passes clean.

jcouball:
Round 7 fix: clarified the `@param value [#to_s]` docs in both the `config(name, value, options)` overload and `config_set` to make clear that `nil` is **not** a valid value — passing `nil` is treated as "no value" and routes to the get overload. Updated wording to "must not be `nil`; any non-nil object is converted to a String via `#to_s`". This is a doc-only change; behavior is unchanged. Rake passes clean.

PR #1516, opened 2026-07-04
build: enforce blank lines between YARD tag groups

jcouball:
## Summary

Enable yard-lint's `Tags/TagGroupSeparator` validator (which is disabled by
default) so the project's house style for blank-line separation between YARD tag
groups is enforced by CI, not just documented in the skills.

Configuration added to `.yard-lint.yml`:

```yaml
Tags/TagGroupSeparator:
  Enabled: true
  Severity: convention
  RequireAfterDescription: true
```

- **Group separation** — a blank line must separate semantically distinct tag
  groups (e.g. `@param` group → `@return` group).
- **`RequireAfterDescription: true`** — a blank line is also required between the
  leading description and the first tag.

## Why

The skills already document this blank-line style, but nothing enforced it. This
closes that gap and pairs naturally with the already-enabled `Tags/Order`
validator.

## Scope / limitation

`Tags/TagGroupSeparator` separates tag **groups** only — it **cannot** require a
blank line between two consecutive same-type tags (e.g. two `@param` tags),
because both map to the same group. Enforcing that stricter per-tag rule needs a
new upstream validator, which is being pursued separately in `mensfeld/yard-lint`.
This PR adopts the group-level enforcement available today as an interim step.

## Baseline

The 23 existing files with `Tags/TagGroupSeparator` offenses (128 offenses total)
are baselined in `.yard-lint-todo.yml`, exactly matching the approach used for the
other validators. They will be fixed incrementally in follow-up PRs, each removing
its file(s) from the todo.

Note: `yard-lint --regenerate-todo` is **not** used here — because `.yard-lint.yml`
inherits from `.yard-lint-todo.yml`, regeneration masks already-excluded offenses
and corrupts the existing baseline. The new section was added manually.

## Verification

- `bundle exec rake default` passes (exit 0): 5668 unit + 893 integration
  examples, rubocop clean, `yard:lint` reports **No offenses found**.

## Checklist

- [x] Tooling/config change on a feature branch targeting `main`
- [x] Conventional Commits format (`build:`)
- [x] No library or spec code changed
- [x] Baseline verified: yard-lint clean, `rake default` green


jcouball:
Closing this PR as superseded. Since it was opened, `main` has adopted `Tags/TagSeparator` (sourced from a fork branch via the Gemfile), which is strictly stronger than `Tags/TagGroupSeparator`: it requires a blank line between **every** consecutive tag, not just between different tag groups. `Tags/TagGroupSeparator` is now explicitly called out in a comment in `.yard-lint.yml` explaining why `Tags/TagSeparator` is used instead. There is nothing left for this PR to add.

PR #1667, opened 2026-08-07
fix: return the full ref path in the ls_remote :ref value

### LINKED ISSUE #1416: ls_remote: :ref value in returned hash is only the first path segment, not the full ref
## Problem

`Git::Base#ls_remote` (and `Git::Repository::RemoteOperations#ls_remote`) returns a Hash where each entry's `:ref` value is only the **first segment** of the full ref path, not the complete ref string.

For example, for a branch at `refs/heads/main` with SHA `abc123`:
```ruby
g.ls_remote('origin')
# => { "branches" => { "main" => { ref: "refs", sha: "abc123" } } }
#                                      ^^^^^ should be "refs/heads/main"
```

The `ref` value is `"refs"` (just the first segment) instead of `"refs/heads/main"` (the full ref path).

## Root cause

In `parse_ls_remote_line`, the full ref string is split on `'/'` to extract the type and name, but only the first segment is stored in `value[:ref]`:

```ruby
sha, info = line.split("\t", 2)  # info = "refs/heads/main"
ref, type, name = info.split('/', 3)  # ref = "refs", type = "heads", name = "main"
value = { ref: ref, sha: sha }  # ref is just "refs", not the full info
```

The fix is to store `info` (the full ref string) instead of `ref` in the value:
```ruby
value = { ref: info, sha: sha }
```

## History

This bug originates in `Git::Lib#ls_remote` on the 4.x branch and was preserved when the facade was introduced so the v5 return value exactly matches v4.x behavior. The fix will require a documented behavior change (the `:ref` value will change from a partial path segment to the full ref string).

## Impact

Callers relying on `result[type][name][:ref]` get an uninformative constant `"refs"` rather than the actual ref path. The `:sha` value is unaffected.

sjh9714:
# Description

Fixes #1416.

`Git::Parsers::LsRemote.parse_line` splits the ref path to derive the type and the
name, but stored only the first segment as `:ref`:

```ruby
sha, info = line.split("\t", 2)       # info = "refs/heads/main"
ref, type, name = info.split('/', 3)  # ref  = "refs"
value = { ref: ref, sha: sha }        # :ref is the constant "refs"
```

So every ref under `refs/` came back with the same uninformative `:ref`:

```ruby
repo.ls_remote('origin')
# before => {"branches"=>{"main"=>{ref: "refs", sha: "abc123"}}}
# after  => {"branches"=>{"main"=>{ref: "refs/heads/main", sha: "abc123"}}}
```

The fix is the one described in the issue: store `info` (the whole ref path). The
first element of the split is then unused, so it becomes `_`.

The `"head"` entry is unaffected. For a `"<sha>\tHEAD"` line, `info` is `"HEAD"` and
the split already yields `"HEAD"`, so its value is identical before and after.
`:sha` is untouched, and so is `parse_default_branch`, which matches the raw output
with its own regex rather than going through `parse_line`.

## Changes

- `lib/git/parsers/ls_remote.rb` — store the full ref path in `:ref`.
- `lib/git/repository/remote_operations.rb` — the two `@example` blocks showed
  `ref: "refs"`; the `@note` documented the truncation and pointed here as the
  planned fix. Both now describe the corrected behavior.
- `spec/unit/git/parsers/ls_remote_spec.rb` — four cases covering the shapes real
  `git ls-remote` emits: branch, tag, a branch name containing a slash, and a
  peeled tag. Each fails on `main` and passes with the fix.
- `spec/unit/git/repository/remote_operations_spec.rb` — assert `:ref` on the parsed
  branch entry; the existing example only asserted `:sha`.
- `spec/unit/git/git_remote_utilities_spec.rb` — one expectation pinned
  `ref: 'refs'`, i.e. the bug itself. Corrected to `'refs/heads/main'`; the test's
  own subject (the execution-context and command wiring) is unchanged.

## A note on versioning

The issue says the fix "will require a documented behavior change", so I want to
flag the release implication rather than decide it. I used a plain `fix:` commit
because 1416 is labeled `bug` and not `breaking-change`, which the repo does apply
to breaking work (issues 1635, 1639 and 1643) — so a plain `fix:` keeps this a
patch bump.
If you would rather it land as a hard break in v6.0.0, say so and I will amend to
`fix!:` with a `BREAKING CHANGE:` footer and add an UPGRADING.md entry. I left
UPGRADING.md alone for now since its sections are per major version.

## Testing

- `LC_ALL=en_US.UTF-8 bundle exec rake` — passes on this branch (exit 0):
  5816 unit examples 0 failures, 569 integration examples 0 failures, RuboCop 621
  files no offenses, yard-lint no offenses, yard example-test 0 failures.
- `bundle exec rake spec:unit` coverage: **100% line (7324/7324) and 100% branch
  (1163/1163)**.
- Same command on unmodified `main` first, for a baseline: exit 0, 5811 examples.
- The four new parser examples were confirmed red before the fix and green after
  (`git stash` on the parser only → 4 failures, all showing `ref: "refs"`).
- Also checked against real `git ls-remote` output through the public API, not just
  mocks: branch, tag and peeled-tag entries all return their full ref path.

One local detail, in case it is useful: I had to set `LC_ALL=en_US.UTF-8` because
`spec/integration/git/command_line_spec.rb:14` matches the English string
`unknown revision`, and git emits localized diagnostics on my machine. That
reproduces on unmodified `main` and is unrelated to this change.

Note: I used Claude Code while preparing this change, and I reviewed the final diff
and ran the listed checks locally.

## Checklist

- [x] `bundle exec rake spec:unit` reports 100% line and branch coverage (see
  [Test coverage policy](../CONTRIBUTING.md#test-coverage-policy)).

### DIFF
--- a/lib/git/parsers/ls_remote.rb
+++ b/lib/git/parsers/ls_remote.rb
@@ -47,12 +47,12 @@ module Git
         end
 
         sha, info = line.split("\t", 2)
-        ref, type, name = info.split('/', 3)
+        _, type, name = info.split('/', 3)
 
         type ||= 'head'
         type = 'branches' if type == 'heads'
 
-        value = { ref: ref, sha: sha }
+        value = { ref: info, sha: sha }
 
         [type, name, value]
       end
--- a/lib/git/repository/remote_operations.rb
+++ b/lib/git/repository/remote_operations.rb
@@ -639,7 +639,7 @@ module Git
       # @example List all refs from the local repository
       #   repo.ls_remote
       #   # => {"head"=>{ref: "HEAD", sha: "abc123"},
-      #   #     "branches"=>{"main"=>{ref: "refs", sha: "abc123"}}}
+      #   #     "branches"=>{"main"=>{ref: "refs/heads/main", sha: "abc123"}}}
       #
       # @example List all refs from a named remote
       #   repo.ls_remote('origin')
@@ -647,7 +647,7 @@ module Git
       #
       # @example List only tags from a named remote
       #   repo.ls_remote('origin', tags: true)
-      #   # => {"tags"=>{"v1.0"=>{ref: "refs", sha: "def456"}}}
+      #   # => {"tags"=>{"v1.0"=>{ref: "refs/tags/v1.0", sha: "def456"}}}
       #
       # @param location [String, nil] the remote name or URL to query; defaults to
       #   `'.'` (the local repository) when nil
@@ -677,12 +677,6 @@ module Git
       # @raise [ArgumentError] if unsupported options are provided
       #
       # @raise [Git::FailedError] if git exits outside the allowed range (exit code > 2)
-      #
-      # @note The `:ref` value in each pair is only the **first path segment** of the
-      #   full git ref (e.g. `"refs"` for `refs/heads/main`), not the complete ref
-      #   path. This matches the behavior of 4.x. See
-      #   [issue 1416](https://github.com/ruby-git/ruby-git/issues/1416) for the
-      #   planned fix.
       #
       def ls_remote(location = nil, opts = {})
         SharedPrivate.assert_valid_opts!(LS_REMOTE_ALLOWED_OPTS, **opts)
--- a/spec/unit/git/git_remote_utilities_spec.rb
+++ b/spec/unit/git/git_remote_utilities_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Git do
 
       expect(described_class.ls_remote('.', tags: true)).to eq(
         'head' => { ref: 'HEAD', sha: 'abc123' },
-        'branches' => { 'main' => { ref: 'refs', sha: 'abc123' } }
+        'branches' => { 'main' => { ref: 'refs/heads/main', sha: 'abc123' } }
       )
     end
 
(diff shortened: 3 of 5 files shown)


jcouball:
@sjh9714 Thanks for addressing #1416. 

I agree with considering this a fix rather than a breaking change. I should have done this before releasing 5.0.0.

I created #1668 to address the issue you reported where you needed to set the locale via LC_ALL when running rake. I'm going to merge that one immediately after this PR and then will release both. Give it a try to let me know if it makes your rake run work without the need to manually set LC_ALL.

sjh9714:
Confirmed — `bundle exec rake` passes without `LC_ALL` now.

This machine is macOS with the system language set to Korean, and git translates its stderr even with `LANG` and `LC_MESSAGES` unset, so the original failure still reproduces here on demand:

- `bundle exec rspec spec/integration/git/command_line_spec.rb` at 34b0b06 (the commit right before your locale pin): 7 examples, **1 failure** — `fatal: 애매한 인자 'nonexistent-ref'` where the example expects `unknown revision`.
- The same spec at v5.0.4 and at current main (c65d402): 7 examples, 0 failures.
- `bundle exec rake` on main with no `LC_ALL` set: green end to end — 5816 unit examples and 569 integration examples (6 pending), 0 failures, RuboCop clean over 621 files, YARD 100% documented, gem builds.

So #1668 does it. Supplying the locale through the spec's own env is the better fix — thanks for picking it up so quickly.

PR #1673, opened 2026-08-07
chore: release v5.0.5

### LINKED ISSUE #1669: Make the pinned LC_ALL platform-conditional (C.UTF-8 on glibc, en_US.UTF-8 on Darwin)
## Summary

`Git::ExecutionContext#env_overrides` pins the locale for every git command:

```ruby
# lib/git/execution_context.rb:456
'LC_ALL' => 'en_US.UTF-8'
```

This proposes replacing that constant with a **platform-conditional** one — `C.UTF-8`
everywhere except Darwin, which keeps `en_US.UTF-8` — documenting *why* the pin exists at
the call site, adding coverage proving the pin actually works, and backporting the same
fix to `4.x`.

The proposal originally read as a robustness improvement. Measurement shows it is a bug
fix: `en_US.UTF-8` is not generated on standard Linux container images, so the pin
silently fails there and every locale-sensitive regex surface in the gem returns wrong
answers today.

```ruby
# after
'LC_ALL' => darwin_platform? ? 'en_US.UTF-8' : 'C.UTF-8'
```

where `darwin_platform?` is
`RUBY_PLATFORM.include?('darwin') || RUBY_DESCRIPTION.include?('darwin')`. `RUBY_PLATFORM`
alone is not enough — see the sixth revision.

## What the evidence shows

**The change is worth making, and — in this form — it costs almost nothing.**

The defect is live and silent on Debian, Ubuntu, and RHEL container images, the
environments where this gem most often runs in production. On eight of the eleven
container images measured, the value pinned today names a locale that does not exist, so
the child process falls back to C ctype and every regex surface in the table below returns
wrong answers with exit status 0.

A flat switch to `C.UTF-8` would fix those and simultaneously **introduce** the identical
defect on macOS 14 and earlier, which have `en_US.UTF-8` but not `C.UTF-8`. An earlier
revision of this issue accepted that trade and proposed a public `Git.config.locale`
escape hatch to compensate. The platform-conditional form avoids the trade instead: each
platform family keeps a value that is known to exist there. macOS is unaffected, because
`en_US.UTF-8` ships on every macOS release and is what the gem already pins.

Both halves of that are measured, not assumed — the benefit on glibc in "Evidence" below,
and the regression that the Darwin branch exists to prevent in "Who actually regresses".

The only platform in the sample that still loses is `centos:7`: glibc 2.17, no `C.UTF-8`,
EOL June 2024, package repositories retired to the point that git could not be installed
on it to measure directly. See "Deferred: the escape hatch" for why that is not being
solved with new public API.

A note on how not to argue this: an earlier revision led with "UTF-8 ctype on 15 of 17
platforms after the change versus 9 today." That number is out (16 of 17 under this
proposal), but it should not be used at all. The seventeen platforms are a convenience
sample chosen for what was easy to run under Docker — with two near-duplicate Debian rows
— not a measure of where the gem is installed. The directional claim survives without it:
this fixes the Linux container substrate and regresses nothing that a supported platform
runs on.

### One mechanism, several symptoms

Every symptom below follows from a single bit: whether the pinned value resolves to a
UTF-8 `LC_CTYPE` in the child process. When it does not — because the named locale is not
generated — glibc falls back to C ctype, the regex engine matches **bytes instead of
characters**, and git's PCRE2 integration stops requesting UTF mode.

Measured on Debian 12 (glibc 2.36, git 2.39.5), Debian 13 (glibc 2.41, git 2.47.3),
Ubuntu 24.04 (glibc 2.39, git 2.43.0) and a macOS 26 developer machine (git 2.55.0),
every row carrying a positive control:

| Surface | Case-sensitive? | Affected? |
| --- | --- | --- |
| `Grep` / `Log` pattern containing `.`, a quantifier, or `{n}` (`^.PFEL`) | **yes** | **yes**, measured |
| `Grep` / `Log` pattern containing a POSIX class or bracket range over non-ASCII (`^[[:alpha:]]PFEL`) | **yes** | **yes**, measured |
| `Grep` / `Log` `perl_regexp` (`-P`) over non-ASCII (`(?i)äpfel`, and `\w` on git ≥ 2.43) | either | **yes**, measured |
| `Grep` — `ignore_case` | no | **yes**, measured |
| `Log` — `regexp_ignore_case` (`--grep`, `--author`, `--committer`) | no | **yes**, measured |
| `Configuring` — `value_regex` on `config_get`, `config_get_all`, `config_get_regexp`, `config_replace_all`, `config_unset`, `config_unset_all`, and the `name_regex` on `config_get_regexp` | yes | **yes**, measured |
| `Commands::Diff` / `Diff::Index` / `Diff::Files` / `Show` — `G`, and `S` with `pickaxe_regex` | yes | **yes**, measured |
| `Commands::Diff` / `Diff::Index` / `Diff::Files` / `Show` — `word_diff_regex` | yes | **yes**, measured — *different symptom*, see below |
| `Grep` — `word_regexp` (`-w`) over non-ASCII | yes | no — measured, unaffected |
| `Branch::List` / `Tag::List` — `ignore_case` | no | no — git's wildmatch folds ASCII only, under every locale |
| paths, content, ref order, message language | — | no — see "Claims that did not survive testing" |

**The table is scoped to glibc, musl, and Darwin.** Windows is a separate regime and the
pinned value does not reach it: measured on `windows-latest` / git 2.55.0 under
`en_US.UTF-8`, `C.UTF-8`, `C`, and no pin at all, the `-i` and `-P` rows work under every
value and the `.` / quantifier / POSIX-class rows fail under every value. So on Windows
nothing in this table is fixed or broken by the pin — the metacharacter surfaces are
already broken there, permanently and independently of this issue. See the Windows bullet
under "Claims that did not survive testing".

The last three rows are **new in this revision** and are the reason the deliverables below
changed. They were found by asking which *other* command classes hand a user-supplied
pattern to git's regex engine, rather than by reasoning forward from `grep`. Two things
about them:

- **They are replicated on two glibc/git pairs**, matching the rigor of the rows above.
  Debian 12 (glibc 2.36, git 2.39.5) and Ubuntu 24.04 (glibc 2.39, git 2.43.0), each in a
  clean container where only `C`, `C.utf8` and `POSIX` are generated, plus the original
  macOS 26 (git 2.55.0) measurement. Full results below. Every probe carried the
  metacharacter-free literal control and every control passed under every value.
- **All of them are, or are becoming, public surface.** The config surfaces are public
  today: `Git::Configuring` is `include`d into `Git::Repository` and `extend`ed onto `Git`,
  so `repo.config_get_all(name, '^.PFEL')` is documented, supported API. The pickaxe and
  word-diff surfaces are not reachable from the facade *yet* — `Git::Repository#diff_full`
  and friends whitelist their options down to `:path_limiter` — but that is not a
  deliberate boundary around these options. The diff whitelists are narrow for two reasons:
  (1) keeping command output parsable, and (2) inherited limits carried over from earlier
  versions. The intent is to keep (1) and remove (2), widening the whitelists to admit as
  many options as possible. So the pickaxe and word-diff rows should be read as *pending*
  public surface, not as private internals — and fixing the pin now means they are correct
  when the whitelists open rather than arriving pre-broken.

#### Measurements for the new rows

Debian 12 and Ubuntu 24.04, each in a clean container. `1` means matched, `0` means did
not match. `ctl` columns are the metacharacter-free literal controls. **PASS 2 repeats the
identical probes after `locale-gen en_US.UTF-8`**, so the only variable between the two
passes is whether the pinned locale exists:

```text
Debian 12 -- glibc 2.36, git 2.39.5    | Ubuntu 24.04 -- glibc 2.39, git 2.43.0
PASS 1: generated = C, C.utf8, POSIX   | PASS 1: generated = C, C.utf8, POSIX
             grep     pickaxe    config|              grep     pickaxe    config
LC_ALL       ctl ^.   ctl -G -S:re ctl gAll gRe        ctl ^.   ctl -G -S:re ctl gAll gRe
en_US.UTF-8  1   0    1   0  0     1   0    0  | en_US.UTF-8  1   0    1   0  0     1   0    0
C.UTF-8      1   1    1   1  1     1   1    1  | C.UTF-8      1   1    1   1  1     1   1    1
C            1   0    1   0  0     1   0    0  | C            1   0    1   0  0     1   0    0

PASS 2: after locale-gen en_US.UTF-8    | PASS 2: after locale-gen en_US.UTF-8
en_US.UTF-8  1   1    1   1  1     1   1    1  | en_US.UTF-8  1   1    1   1  1     1   1    1
C.UTF-8      1   1    1   1  1     1   1    1  | C.UTF-8      1   1    1   1  1     1   1    1
C            1   0    1   0  0     1   0    0  | C            1   0    1   0  0     1   0    0
```

Under the value pinned **today**, a real glibc host answers every metacharacter probe
exactly as `LC_ALL=C` does — because on these hosts it *is* an absent locale. Generating
the locale flips every zero to a one without touching git, the repository, or the pattern.

#### `word_diff_regex` fails differently, and the first probe for it was wrong

`--word-diff-regex` is a tokenizer, not a matcher, so it has no "did it match" bit. The
first attempt counted changed tokens — that count is **2 under every locale**, so it
discriminated nothing and should not be reused. The signal is token *content*: under UTF-8
ctype the changed token is a whole character; under C ctype the shared lead byte is emitted
as an *unchanged* token and only the trailing byte differs, so the emitted tokens are
fragments of a character and are not valid UTF-8.

Diffing `ÄPFEL` (C3 84) against `ÜPFEL` (C3 9C) with `--word-diff-regex='.'`, showing the
raw bytes of the changed-token lines (`2d` = `-`, `2b` = `+`, `0a` = newline):

```text
Debian 12, PASS 1 (en_US.UTF-8 absent)          Ubuntu 24.04, PASS 1
  en_US.UTF-8  INVALID -- split   2d 84 0a 2b 9c 0a        same
  C.UTF-8      VALID              2d c3 84 0a 2b c3 9c 0a  same
  C            INVALID -- split   2d 84 0a 2b 9c 0a        same

PASS 2 (after locale-gen), both platforms
  en_US.UTF-8  VALID              2d c3 84 0a 2b c3 9c 0a
  C.UTF-8      VALID              2d c3 84 0a 2b c3 9c 0a
  C            INVALID -- split   2d 84 0a 2b 9c 0a
```

A silently empty result set is one failure class; output carrying invalid UTF-8 that a
caller will then try to encode is another, and it is likelier to surface as a loud
downstream error than as a wrong answer.

#### End-to-end through the public API

The `grep` reproduction below is joined by the config surface, since that one is public
today. `ruby:3.4-slim` (ruby 3.4.10, git 2.47.3, glibc 2.41; generated locales `C`,
`C.utf8`, `POSIX`), same process and repository, only the pinned value differing:

```console
$ # pin as shipped: 'LC_ALL' => 'en_US.UTF-8'
repo.config_get_all('test.desc', 'ÄPFEL')    # control
=> [#<data Git::ConfigEntryInfo ... value="ÄPFEL sind gut">]
repo.config_get_all('test.desc', '^.PFEL')
=> []
repo.grep('äpfel', nil, ignore_case: true)
=> {}

$ # pin patched to: 'LC_ALL' => 'C.UTF-8'
repo.config_get_all('test.desc', 'ÄPFEL')    # control
=> [#<data Git::ConfigEntryInfo ... value="ÄPFEL sind gut">]
repo.config_get_all('test.desc', '^.PFEL')
=> [#<data Git::ConfigEntryInfo ... value="ÄPFEL sind gut">]
repo.grep('äpfel', nil, ignore_case: true)
=> {"HEAD:w.txt" => [[1, "ÄPFEL sind gut"]]}
```

The control returns the same entry under both values, so the empty array is a real result.

A silently empty result set is one failure class; output containing broken UTF-8 that a
caller will then try to encode is another, and it is the one more likely to surface as a
loud downstream error rather than a wrong answer.

**This is a correction to an earlier revision of this issue, which claimed the defect was
confined to the two case-insensitive surfaces.** It is not. The earlier positive control
was a metacharacter-free literal (`git grep 'ÄPFEL'`), which passes under C ctype because
a literal byte sequence matches byte-for-byte. A control of that shape is structurally
incapable of detecting a byte-vs-character defect, and the "case-sensitive matching is
fine" conclusion was an over-generalization from it.

The practical consequence is that the exposure is **wider than previously stated**: it
requires a glibc host without `en_US.UTF-8` **and** a pattern containing any regex
metacharacter that spans non-ASCII text. It no longer requires a case-insensitive search.
Severity per occurrence is unchanged and high — a wrong answer with exit status 0.

That also widens the regression on macOS ≤ 14 by exactly the same amount, and it is why
the CHANGELOG entry has to describe the change as "non-ASCII regex matching" rather than
"case-insensitive search".

It remains true that no field report exists in the ~18 months since the pin landed. This
repository has eight closed non-ASCII issues — #418, #188, #128, #33, #313, #308, #295,
and #387 — and every one is a *loud* failure: a crash or garbled output. A silently empty
result set is precisely the failure mode that does not get reported, so the silence is
weak evidence either way; the honest summary is high severity, low frequency, and a
frequency somewhat higher than the previous revision estimated.

`lib/git/execution_context.rb:456` is the only place the gem sets this environment;
nothing under `lib/` shells out to git outside `Git::CommandLine`, and nothing else calls
`env_overrides`, so the pin is a single choke point and the fix is complete.

## Background: why the pin exists

The pin was added by `f407b92d` ("feat: set the locale to en_US.UTF-8 for git commands",
2025-02-27) in response to #753, where a German user's `git branch` output came back as
`* (HEAD losgelöst bei origin/25.1)` and broke branch parsing.

Three things are worth noting:

1. The commit has an **empty body** and there is no comment at the call site, so the
   reasoning was never recorded. Nothing currently tells a future contributor why the
   value is what it is.
2. The reporter's own workaround in that thread was `LC_ALL=C` — a value that is subtly
   wrong for reasons documented below. That is exactly the change someone might make while
   "cleaning up" the line.
3. **The stated rationale is now stale.** The pin was justified as protecting parsers from
   translated text, but in v5 `lib/git/repository/branching.rb:224` and `:245` read
   `--format=%(refname:short)` rather than `git branch`'s human-readable output, and no
   code under `lib/` matches an English git message. The exact #753 failure mode is
   structurally fixed independent of the locale. The message half of the pin is now
   defense-in-depth for user code and for the remaining human-readable surfaces; the
   **ctype half is the half doing load-bearing work**, which is the opposite of how the
   value reads today.

### Method

Eleven container images under Docker, six GitHub-hosted runners, and one developer machine.
For each: inventory the generated locales, then measure behavior functionally rather than
inferring it from `locale -a`.

- **ctype** — `git grep` and `git log --grep` for the patterns in the surface table above,
  against a file and commit message containing `ÄPFEL sind gut`, under each candidate
  `LC_ALL`.
- **Message language** — `git rev-parse nonexistent-ref` under each candidate.
- **perl stderr** — line count on stderr from a trivial `perl -e` under each candidate.

Container images were probed on arm64 except `centos:7` (amd64); locale databases are
architecture-independent. git versions ranged from 2.25.1 to 2.55.0 across the sample, so
the results are not specific to one build's regex or gettext configuration.

Every row carries a **positive control**: the same pattern matched as a metacharacter-free
literal (`git grep 'ÄPFEL'`) must succeed under every locale. Without it a broken test rig
is indistinguishable from a real no-match result — an early run of these probes reported a
spurious no-fold for `C.UTF-8` because the pattern reached `git grep` as an uninterpreted
escape sequence rather than as UTF-8 bytes. Note the limit of this control, learned the
hard way: it proves the rig delivers UTF-8 bytes, and proves nothing about whether
case-sensitive *pattern* matching is intact (see the correction above).

### Platform results

The `ctype` columns record whether the pinned value resolves to a UTF-8 `LC_CTYPE`, which
is the single bit every symptom follows from.

The last two columns are the decision. "flat `C.UTF-8`" is the rejected alternative — one
constant everywhere — kept so the cost it would have carried stays visible.

| Platform | libc / runtime | `C.UTF-8` | `en_US.UTF-8` | ctype: as shipped | ctype: flat `C.UTF-8` *(rejected)* | ctype: **this proposal** |
| --- | --- | --- | --- | --- | --- | --- |
| `ruby:3.4-alpine` (3.24) | musl | n/a | n/a | UTF-8 | UTF-8 | UTF-8 |
| `ruby:3.4-slim` (Debian 13) | glibc 2.41 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `debian:trixie-slim` (13) | glibc 2.41, git 2.47.3 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `debian:bookworm-slim` (12) | glibc 2.36, git 2.39.5 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `debian:bullseye-slim` (11) | glibc 2.31 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `ubuntu:20.04` | glibc 2.31 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `ubuntu:24.04` | glibc 2.39, git 2.43.0 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `almalinux:9` (RHEL 9) | glibc 2.34 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `almalinux:8` (RHEL 8) | glibc 2.28 | yes | **no** | **C — broken** | UTF-8 | UTF-8 |
| `amazonlinux:2` | glibc 2.26 | yes | yes | UTF-8 | UTF-8 | UTF-8 |
| `centos:7` (RHEL 7) | glibc 2.17 | **no** | yes | UTF-8 | **C — regresses** | **C — regresses** |
| `ubuntu-latest` runner | glibc, git 2.54.0 | yes | yes | UTF-8 | UTF-8 | UTF-8 |
| `windows-latest` runner | MinGW64, git 2.55.0 | yes | yes | UTF-8 | UTF-8 | UTF-8 |
| `windows-2022` runner | MinGW64, git 2.55.0 | yes | yes | UTF-8 | UTF-8 | UTF-8 |
| `macos-14` runner | Darwin 23.6 | **no** | yes | UTF-8 | **C — regresses** | UTF-8 — *keeps `en_US.UTF-8`* |
| `macos-15` runner | Darwin 24.6 | yes | yes | UTF-8 | UTF-8 | UTF-8 — *keeps `en_US.UTF-8`* |
| `macos-26` runner | Darwin 25.5 | yes | yes | UTF-8 | UTF-8 | UTF-8 — *keeps `en_US.UTF-8`* |

Message language was English under both candidate values on every platform tested, so the
pin's original purpose — keeping translated text away from the parsers — is satisfied
either way.

On `centos:7` the ctype column is derived from `setlocale` rather than from `git grep`: its
package repositories are retired, so git could not be installed. `LC_ALL=C.UTF-8 locale
charmap` fails there with `Cannot set LC_CTYPE to default locale` and falls back to
`ANSI_X3.4-1968`, which is the same mechanism measured directly on the other sixteen
platforms.

### The defect is not confined to case-insensitive search

Debian 12, glibc 2.36, **git 2.39.5**, with `en_US.UTF-8` genuinely absent (`C C.utf8
POSIX` are the only generated locales). `1` means matched, `0` means did not match:

```text
LC_ALL       control  grep ^.  grep [:alpha:]  log ^.  -P (?i)  grep -i
en_US.UTF-8  1        0        0               0       0        0
C.UTF-8      1        1        1               1       1        1
C            1        0        0               0       0        0
xx_YY.UTF-8  1        0        0               0       0        0
```

Replicated on Debian 13, glibc 2.41, **git 2.47.3**, same result, and on Ubuntu 24.04,
glibc 2.39, **git 2.43.0**. On a macOS 26 developer machine (Darwin 25.6, git 2.55.0) the
same divergence appears between `LC_ALL=C` and `LC_ALL=en_US.UTF-8`, which is the same
mechanism reached by a different route.

Four of those six probes are **case-sensitive**. `en_US.UTF-8` behaves identically to `C`
and to a locale that does not exist, because on these hosts it *is* a locale that does not
exist. The `control` column passes on every row, so these are real no-match results.

`-P '\w'` folds under `C.UTF-8` on git 2.43.0 and 2.47.3 but not on 2.39.5, so the `\w`
case specifically is git-version dependent; `-P '(?i)…'` is affected on every version
tested.

### Effect through the gem

Reproduced on `ruby:3.4-slim` (ruby 3.4.10, git 2.47.3), same process and same repository,
with only the pinned value differing:

```console
$ # locales generated in this image: C, C.utf8, POSIX
$ # pin as shipped: 'LC_ALL' => 'en_US.UTF-8'
repo.grep("äpfel", nil, ignore_case: true)
=> {}

$ # pin patched to: 'LC_ALL' => 'C.UTF-8'
repo.grep("äpfel", nil, ignore_case: true)
=> {"HEAD:w.txt" => [[1, "ÄPFEL sind gut"]]}
```

An empty hash, no exception, and no warning. A wrong answer with exit status 0 is worse
than a noisy failure, and this is the user-visible consequence of the pin failing. The
`control` run — the same pattern matched case-sensitively — passes under both values, so
the empty hash is a real result and not a broken harness.

### Where both locales exist, the two values really are interchangeable

The "indistinguishable" claim initially rested on three probes (folding, message language,
perl stderr), none of which reach collation. POSIX bracket ranges are collation-sensitive
in principle — under some libc implementations `[a-z]` matches uppercase in a non-C locale
— so this was checked directly. After `locale-gen en_US.UTF-8` on the same image, with
both locales real:

```text
en_US.UTF-8  grep-i:FOLDS  log-i:FOLDS  [a-z] vs "ONLYUPPER":no-match  alpha/upper/punct: 1/1/0  tags: Alpha apple zeta
C.UTF-8      grep-i:FOLDS  log-i:FOLDS  [a-z] vs "ONLYUPPER":no-match  alpha/upper/punct: 1/1/0  tags: Alpha apple zeta
```

Identical on every axis. Modern glibc resolves range expressions by codepoint rather than
by collation weight, so the hypothesized divergence does not occur. Re-confirmed on
Debian 13 and Ubuntu 24.04 across the six-probe set above, which adds the case-sensitive
metacharacter axes: with both locales generated, every probe returns `1` under both values.

### perl warnings: real, but not part of the argument

Every image where `en_US.UTF-8` is absent emits 7 to 19 lines of
`perl: warning: Setting locale failed.` on stderr **today**, and zero under `C.UTF-8`.

This is **not load-bearing and should not be cited as a reason to make the change.** No
class under `lib/git/commands/` shells out to a perl-backed subcommand, and
`lib/git/command_line/base.rb:242` raises only on a non-zero exit status — stderr content
alone fails nothing. The exposure is limited to noise in a captured stderr string when a
user's own perl hook runs. Recorded as a side benefit; the case for the change rests on
the ctype defect alone.

### Claims that did not survive testing

Recorded so they are not rediscovered:

- **The defect is confined to case-insensitive matching.** Withdrawn — see the correction
  above. It affects any pattern whose metacharacters span non-ASCII text, case-sensitive
  or not.
- **Plain `C` silently changes results.** `LC_CTYPE=C` matches bytes rather than
  characters, so non-ASCII patterns quietly return wrong answers. This is the trap the
  call-site comment must guard against.
- **A missing locale does not degrade entirely safely.** git falls back to English
  messages, which is the goal — but the process also falls back to C ctype. Both halves of
  the value fail together, and only one of the two failures is benign.
- **Alpine is not a beneficiary.** musl ignores the locale name for ctype purposes:
  matching works there under `en_US.UTF-8` and even under a nonexistent `xx_YY.UTF-8`. The
  practical beneficiaries are Debian, Ubuntu, and RHEL images.
- **Windows is unaffected *by the pinned value* — but is not otherwise fine.** Measured on
  `windows-latest`, git 2.55.0, `x64-mingw-ucrt`, under `en_US.UTF-8`, `C.UTF-8`, `C`, and
  no pin at all: every probe returns the same answer under all four. Folding (`-i`) and
  `-P` work under all of them; `.` and POSIX classes over non-ASCII match bytes under all
  of them. Git for Windows is locale-independent here, so the pin neither helps nor harms
  it.

  This corrects an earlier revision, which read "Windows is unaffected. Git for Windows
  folds non-ASCII under every value tested." Every word of that was true and its
  conclusion for the pin was right, but it rested on a *folding* probe, and the third
  revision widened the surface table to metacharacter patterns without re-running Windows.
  Note the shape of the error: it is the third time in this issue's history that a probe
  of one shape has been generalized to a surface it structurally cannot measure.

  The practical consequence is separate from this issue and worse than "unaffected"
  suggests: non-ASCII metacharacter matching does not work on Git for Windows at all, in
  `grep`, `log`, or the config value regexes. `-P` does work there, so `perl_regexp` may be
  a viable workaround — though `Git::Repository#grep` does not expose it today
  (`GREP_ALLOWED_OPTS` is `ignore_case`, `invert_match`, `extended_regexp`, `object`).
  Worth its own issue.
- **`C.UTF-8` is far more widely available than upstream glibc history suggests.**
  Distributions backported it well before glibc 2.35: it is present on glibc 2.26 (Amazon
  Linux 2), 2.28 (RHEL 8), 2.31 (Debian 11, Ubuntu 20.04), and 2.34 (RHEL 9). On macOS it
  arrived in macOS 15 — absent on Darwin 23.6, present on Darwin 24.6.
- **Paths and commit content are locale-independent.** `C` and `en_US.UTF-8` produce
  byte-identical output, both with `core.quotePath=true` (octal-escaped) and with it off.
  A commit by `José Müller` with subject `añadir soporte für Ünicode — 日本語` diffs
  identically; content encoding follows `i18n.logOutputEncoding` / `i18n.commitEncoding`,
  not the locale charset.
- **Ref ordering is not collation-sensitive.** git compares refs bytewise, so `git branch`
  output order is identical under both values.
- **Regex bracket *ranges* are not collation-sensitive — hypothesis tested and withdrawn.**
  POSIX says range expressions like `[a-z]` may follow collation order, which would have
  made the two values differ on hosts where both locales exist. Measured on glibc with both
  generated: identical. Modern glibc resolves ranges by codepoint. Note this is a separate
  question from whether a bracket expression *containing non-ASCII characters* works at
  all, which is ctype-dependent and does break — see the surface table.
- **`git tag --list -i` and `git branch --list -i` do not fold non-ASCII under any locale.**
  git's wildmatch case-folding is ASCII-only, so these two surfaces are neither broken
  today nor improved by the change.
- **`git grep -w` is ctype-dependent.** Measured unaffected: `-w ÄPFEL` matches under every
  value including `C`.
- **`LC_TIME` / `LC_NUMERIC` exposure matters.** `LC_ALL` overwrites every category, not
  just ctype and messages, so the change does alter time and number formatting for git and
  for every process it spawns. Nothing under `lib/` parses locale-formatted dates or
  numbers (git formats its own dates; no `strftime`, `Time.parse`, or `DateTime.parse` on
  command output), and both `en_US` and `C` are English, so no gem-visible difference. Kept
  on the list because it is the one axis where the single-variable `LC_ALL` form is
  measurably broader than the alternative below, and because it reaches user hooks.

## Why the Darwin branch exists

Without it, this change would break macOS ≤ 14 exactly as thoroughly as it fixes Debian.
That claim used to be **inferred** from the availability table — `C.UTF-8` is absent on
Darwin 23.6, therefore the pin fails, therefore C ctype — which assumes BSD libc degrades
the way glibc does. That last step had never been checked. It does. Measured on macOS 26 /
git 2.55.0 with a nonexistent locale standing in for `C.UTF-8` on macOS ≤ 14, controls
passing:

```text
LC_ALL         ctl    grep ^.   config value_regex   charmap
en_US.UTF-8    1      1         1                    UTF-8
C.UTF-8        1      1         1                    UTF-8
xx_YY.UTF-8    1      0         0                    US-ASCII
C              1      0         0                    US-ASCII
```

An absent locale on BSD libc falls back to US-ASCII ctype and behaves identically to `C` —
the same single mechanism as glibc, reached on a different libc. So a flat `C.UTF-8` would
produce every failure in the surface table on macOS ≤ 14, and the population is larger than
"one superseded release":

- `C.UTF-8` arrived in **macOS 15**, so macOS 11, 12, 13, and 14 all lack it — every macOS
  released before September 2024.
- A slice of Intel Macs is **hardware-capped** below macOS 15 (2017 MacBook Pro and iMac,
  2019 MacBook Air and earlier). Those machines will never gain `C.UTF-8`.
- The [Ruby Version Support Policy](README.md) promises this gem "is expected to function
  correctly on all non-EOL versions of MRI Ruby on Mac, Linux, and Windows" with **no macOS
  floor**. Ruby 3.2 (the gemspec minimum) runs fine on macOS 12, so a flat value would
  knowingly degrade an in-policy platform.

The Darwin branch removes all of that. `en_US.UTF-8` is part of the macOS base system on
every release measured (Darwin 23.6, 24.6, 25.5 all have it), and it is what the gem pins
today — so macOS keeps its current, working behavior and this proposal has no macOS
regression to disclose, gate in CI, or compensate for.

Test the platform plainly, **not** by Darwin version. `RUBY_PLATFORM` records the version
Ruby was *built* against, so a Ruby built on macOS 14 and run on macOS 15 still reports
`darwin23`. The plain platform test does not depend on that; version sniffing would be
wrong. The branch is dead weight once macOS ≤ 14 is gone, which is harmless and should be
noted in the call-site comment.

**`RUBY_PLATFORM` alone does not make that test correctly**, which earlier revisions of
this issue got wrong. JRuby reports it as `"java"` on every host and records the real
platform only in `RUBY_DESCRIPTION`, so the detection has to be
`RUBY_PLATFORM.include?('darwin') || RUBY_DESCRIPTION.include?('darwin')`. See the sixth
revision.

## Deferred: the escape hatch

There is no way for a user to override the pin. This is still true, and it is worth
recording precisely, because a previous revision made it a deliverable:

- `env_overrides` is `@api private`, and **nothing under `lib/` passes
  `additional_overrides`** — that parameter is exercised only by specs.
- `Git::ExecutionContext#initialize` (line 110) accepts `binary_path:`, `git_ssh:`, and
  `logger:`. There is no `env:` or `locale:`.
- `Git::Config` exposes `binary_path`, `git_ssh`, and `timeout` writers. There is no
  locale setting.
- The override wins over the parent environment, so exporting `LC_ALL=en_US.UTF-8` in the
  user's shell has no effect on the child.

A previous revision proposed adding `Git.config.locale` plus a `locale:` constructor
keyword to compensate for the macOS regression. **Decision: deferred, not rejected.** The
reasoning:

1. **Its main justification is gone.** It existed to give macOS ≤ 14 users a way back to a
   working value. Under this proposal they never leave one.
2. **The remaining population is `centos:7` and untested Unixes** (FreeBSD, Solaris, AIX) —
   a distribution EOL since June 2024 with retired package repositories, plus platforms
   with no reported users and no CI coverage. That does not justify permanent public API.
3. **It would not have helped the people it was for.** The defect is silent, so a user only
   reaches for an override *after* diagnosing it. The hatch serves the already-diagnosed,
   which is a much narrower job than it appears.
4. **It conflicts with the rationale it was invoked to defend.** `Git::Config` is a
   process-wide mutable singleton, so any dependency could change git's locale for the
   whole host application — a nondeterminism knob added in service of the determinism goal
   stated for `STATIC_GLOBAL_OPTS`.
5. **It does not fit the pattern it claimed to copy.** `binary_path`, `git_ssh`, and
   `timeout` all initialize to `nil` and read as `@x || fallback`, where `nil` means
   "unconfigured, use the default". The hatch needs a *third* state — pin the default, pin
   a supplied value, or **pin nothing at all** — and `nil` cannot mean both "unconfigured"
   and "do not pin". It would need an explicit sentinel (`@locale = :use_default` in
   `initialize`, `nil` meaning omit the key), which is a deviation from the existing idiom
   rather than an instance of it. Anyone reviving this should design that explicitly, and
   should also decide whether the writer validates — setting a nonexistent locale silently
   yields C ctype, which is the exact trap the call-site comment warns about.

Revisit if someone on RHEL 7 or an untested Unix reports a real problem. Until then the
call-site comment carries the knowledge instead.

## Proposal

Make the pinned value platform-conditional:

```ruby
'LC_ALL' => darwin_platform? ? 'en_US.UTF-8' : 'C.UTF-8'

def darwin_platform?
  RUBY_PLATFORM.include?('darwin') || RUBY_DESCRIPTION.include?('darwin')
end
```

Each branch names a locale that is known to exist on the platform it applies to. On glibc
and musl hosts `C.UTF-8` gives English messages and correct non-ASCII regex matching, and
it says what it means ("no localization, UTF-8 charset") rather than impersonating an
American locale. On Darwin, `en_US.UTF-8` is part of the base system on every release and
is already what the gem pins, so nothing changes there. (Silencing the perl warnings on
glibc is a side effect, not a justification.)

This is a constant per platform, not a runtime probe — see "Acceptance criteria and
non-goals", which rules out probing and which this satisfies: no subprocess, no `locale -a`,
no filesystem inspection, and trivially stubbed in a unit test.

### Deliverables

Ship in three parts. Part 1 fixes users' problems on `main` and depends on nothing; Parts 2
and 3 are independent of each other and of Part 1's review cycle.

**Part 1 — the fix** — done in #1681.

- [x] Replace the constant in `lib/git/execution_context.rb:456` with the
      platform-conditional form: `RUBY_PLATFORM.include?('darwin') ? 'en_US.UTF-8' :
      'C.UTF-8'`. Use the plain `darwin` test, not a Darwin version check — see "Why the
      Darwin branch exists". (Corrected in the sixth revision: `RUBY_PLATFORM` alone
      misses JRuby. Shipped as specified in #1681 and fixed in #1684.)
- [x] Add a comment at the call site recording *why* the pin exists (cite #753), that the
      **ctype** half is the load-bearing half now that `branching.rb` reads `--format`,
      why plain `C` is not acceptable, why the Darwin branch exists (`C.UTF-8` arrived in
      macOS 15) and that it can be deleted once macOS ≤ 14 is out of support, and that
      RHEL 7 has neither value. This is the part that survives the next contributor.
- [x] Update the `LC_ALL=en_US.UTF-8` reference in `CONTRIBUTING.md:304` — it now needs to
      describe a platform-conditional value rather than a literal one.
- [x] Update the ten existing assertions on the literal value, across four spec files:
      `spec/unit/git/execution_context_spec.rb` (5), `execution_context/repository_spec.rb`
      (2), `execution_context/global_spec.rb` (2), and the constructor argument in
      `spec/integration/git/command_line_spec.rb:16` (1). Some are example descriptions
      naming the value, not just `hash_including` matchers. **These can no longer assert a
      bare literal** — they must either derive the expected value the same way the source
      does or stub `RUBY_PLATFORM`, or they will pass on Linux and fail on macOS. Add one
      example per branch so both are covered on every host.
- [x] ~~Replicate the three new rows of the surface table on a real glibc host with
      `en_US.UTF-8` genuinely absent.~~ Done — Debian 12 and Ubuntu 24.04, both passes,
      results above. The public blast radius is `grep`, `log`, and the six `Configuring`
      methods that take a `value_regex`, with the diff surfaces joining it as the option
      whitelists widen.
- [x] Add coverage that non-ASCII regex matching works through every **public** affected
      surface — `Git::Repository#grep`, the `log` path, and at least one `Configuring`
      `value_regex` method — and for **both** a case-sensitive metacharacter pattern
      (`^.PFEL`) and `ignore_case`. Covering only `ignore_case`, or only `grep` and `log`,
      leaves most of the measured defect unguarded. See the constraint below.
      Added as `spec/integration/git/non_ascii_regex_matching_spec.rb`, with the three
      metacharacter examples skipped on Windows — see the constraint below for why.
- [x] Add a CHANGELOG entry. Describe the effect as "non-ASCII regex matching in `grep`,
      `log`, and `config` value regexes", not "case-insensitive search" and not "in `grep`
      and `log`". Unlike the previous revision, this entry has **no losing platform to
      disclose** — it is a fix on glibc hosts and a no-op everywhere else.
      `CHANGELOG.md` is generated from conventional commits, so the commit message is
      the entry; nothing is hand-edited.

**Part 2 — the 4.x backport** — done in #1682.

- [x] Apply the same platform-conditional value at `lib/git/lib.rb:2091` on the `4.x`
      branch, with the same call-site comment. See "Scope".
      The comment differs in one respect, deliberately: on `4.x`,
      `Git::Lib::BRANCH_LINE_REGEXP` still matches the English `(HEAD detached at ...)`
      and `(not a branch)` strings in `git branch -a` output — the exact #753 failure
      mode — so **both** halves of the pin are load-bearing there, not just the ctype
      half as on `main`. The backport also carries the JRuby detection fix described in
      the sixth revision, updates the four tests that hardcoded the literal, and adds
      `grep` and `log` regression coverage. The six `Configuring` `value_regex` methods
      do not exist on `4.x`, so that surface is complete rather than reduced.

**Part 3 — the regression guard** (separate PR; protects against a *future* wrong value)

- [x] Add the discriminating spec run described below.

### Constraint on the new coverage

A matching assertion is a statement about the host's locale database, not about ruby-git,
and the matrix cannot currently detect a regression in it: on `ubuntu-latest` both locales
exist, so the assertion passes under **either** value and has no discriminating power.

The previous revision also needed a Darwin skip-gate here, because a flat `C.UTF-8` would
have turned macOS ≤ 14 checkouts red. **That constraint is gone** — the Darwin branch keeps
the value that works there, so no Darwin gate is needed.

Windows does need one, which was not known when this was written. Git for Windows matches
bytes for `.` and POSIX classes over non-ASCII text identically under every value
(measured; see the Windows bullet above), so the metacharacter examples fail there under
the value shipped today just as they do under `C.UTF-8`. They are skipped on
`Gem.win_platform?`, with the measurement recorded at the skip site. That gate reads the
platform and never the locale, so it does not weaken the rule below. The `ignore_case`
example and every positive control run unguarded on all platforms.

One thing to preserve from that discussion: never gate the spec on *probing whether the
pinned locale loads*. A spec that skips whenever the pinned locale is missing disappears in
exactly the environments where the bug lives, which is how this defect survived since
February 2025.

- **Make one run discriminating**, using `LOCPATH` rather than a container image. `C.UTF-8`
  is compiled into glibc ≥ 2.35 and needs no locale archive; `en_US.UTF-8` always needs
  one. Pointing `LOCPATH` at an empty directory therefore removes `en_US.UTF-8` while
  leaving `C.UTF-8` working. **Measured on `ubuntu:24.04`, glibc 2.39 — the CI runner's
  base — with `en_US.UTF-8` generated first:**

  ```text
  normal        en_US.UTF-8 charmap=UTF-8            C.UTF-8 charmap=UTF-8
  LOCPATH=empty en_US.UTF-8 charmap=ANSI_X3.4-1968   C.UTF-8 charmap=UTF-8
  ```

  All six functional probes pass under `C.UTF-8` and fail under `en_US.UTF-8` in the
  `LOCPATH` run. So a discriminating guard is two lines in an existing ubuntu job — no
  container image, no separate `bundle install`, no lost bundler cache. Verify it still
  discriminates when adopted; the fallback is
  `sudo localedef --delete-from-archive en_US.UTF-8`, still cheaper than a container job.
  (`locale -a` continues to list `en_US.utf8` under `LOCPATH`, so the guard must assert
  functionally, not from `locale -a`.)

### CI cannot see macOS at all — which is now an argument *for* this form

The build matrix is four Rubies on `ubuntu-latest` plus one `windows-latest` job, with
separate lint and non-English-locale jobs (`.github/workflows/continuous_integration.yml`).
**There is no macOS job.**

The previous revision treated this as a liability: a flat `C.UTF-8` would have degraded
macOS silently, and no CI job would have caught it — the failure mode being not a red build
but a wrong answer on a contributor's machine, which is the same failure class this issue
exists to fix, relocated rather than eliminated.

Under the platform-conditional form this inverts. The Darwin branch is the one platform
family CI cannot verify, and it is also the branch that **changes nothing** — it pins the
value macOS already uses. The untested path is the unchanged path. That is a deliberate
part of why this form was chosen over a flat constant, and it is worth keeping in mind if
anyone later proposes "simplifying" the ternary away: doing so would silently degrade the
exact platform the matrix cannot see.

The blind spot is not hypothetical. It swallowed a real defect: the JRuby detection bug in
the sixth revision put JRuby on macOS onto the wrong branch, and no job in either matrix
could have caught it, because there is no macOS job and JRuby runs only on
`ubuntu-latest`. It was found by reading, not by CI.

### Acceptance criteria and non-goals

- Availability is confirmed; see the table above. No further platform research is required
  before acting.
- **Do not build a locale-probing fallback** (try `C.UTF-8`, fall back to `en_US.UTF-8`,
  fall back to `C`). Note that the platform-conditional constant this issue proposes is
  **not** such a fallback — it is resolved from `RUBY_PLATFORM` at load time, with no
  subprocess, no `locale -a`, no filesystem inspection, and no dependence on the host's
  locale database. It is a constant per platform and is stubbable in a unit test. The
  objections below apply to *runtime probing*, which remains ruled out:

  1. A probe cannot be opportunistic, because a failed pin does not fall back to the
     ambient locale — it falls back to C. Measured:

     ```text
     LANG=en_US.UTF-8  LC_ALL=<nonexistent>                        C ctype
     LANG=en_US.UTF-8  LC_CTYPE=<nonexistent> LC_MESSAGES=C        C ctype
     LANG=en_US.UTF-8  (nothing pinned)                            UTF-8 ctype
     ```

     So the fallback would have to genuinely enumerate the available locales.
  2. From Ruby that means shelling out to `locale -a` (an extra subprocess, memoized, and
     absent on Windows and minimal images) or inspecting a `/usr/lib/locale` path that
     varies by distribution. The `LOCPATH` result above shows `locale -a` is not even
     reliable for this: it lists locales that `setlocale` then refuses. That is a poor
     trade in a gem whose value is being a thin, predictable wrapper around the CLI.
  3. Host-dependent environment construction is also materially harder to test than a
     constant, and would need its own matrix to be trustworthy.

  Most of what a probe would buy is bought instead by the platform branch, at compile-time
  cost and with none of the testing burden.

  The consequence is that the pin still cannot guarantee identical behavior *everywhere* —
  a pinned locale that does not exist yields C ctype, as the `centos:7` row shows. This
  proposal is better because each branch names a locale that is known to exist on the
  platform it applies to, not because pinning guarantees uniformity.

## Alternative considered: a single flat `C.UTF-8` plus a public escape hatch

This was the proposal through the first three revisions of this issue, and it is the
obvious reading of the title. **Rejected in favor of the platform branch.**

It changes the constant to `C.UTF-8` on every platform, accepts that macOS ≤ 14 and RHEL 7
lose non-ASCII regex matching as a result, and compensates by adding `Git.config.locale`
plus a `locale:` constructor keyword so affected users can pin something else.

What it costs, all of which the platform branch avoids:

- **A measured regression on an in-policy platform.** Every failure in the surface table
  would newly occur on macOS 11 through 14, including hardware-capped Intel Macs that can
  never upgrade. See "Why the Darwin branch exists".
- **Permanent public API** — a locale setter on a process-wide mutable singleton — carrying
  a three-state design that does not fit the `binary_path` idiom it claimed to follow. See
  "Deferred: the escape hatch".
- **A compensation that does not reach the affected.** The defect is silent, so users only
  reach for an override after diagnosing it.
- **A CHANGELOG entry that must name losing platforms**, and a Darwin skip-gate in the new
  spec, and an argument about whether the support policy needs an explicit macOS floor.

What it buys over the platform branch: one constant instead of two, and a supported way for
anyone on any platform to override the pin. The second of those is real, and is the reason
the escape hatch is recorded as deferred rather than rejected outright.

## Alternative considered: force only the message language

Set `LC_MESSAGES` to `C` and leave `LC_ALL` unset (`env_overrides` already documents `nil`
as "unset"). Verified:

```console
$ LANG=de_DE.UTF-8                git rev-parse nonexistent-ref
Schwerwiegend: mehrdeutiges Argument 'nonexistent-ref': ...

$ LANG=de_DE.UTF-8 LC_MESSAGES=C  git rev-parse nonexistent-ref
fatal: ambiguous argument 'nonexistent-ref': ...        # and non-ASCII matching still works
```

Technically the most surgical option — English guaranteed via a locale POSIX mandates
exists, charset inherited from the user. **Rejected** because it conflicts with the goal
stated for `STATIC_GLOBAL_OPTS`, "deterministic, script-friendly output regardless of the
user's local git configuration": inheriting `LC_CTYPE` means matching depends on each
machine's ambient locale, and in a container with no `LANG` set that means C ctype and no
non-ASCII matching at all — the defect this issue exists to fix. It is, however, exactly
what `Git.config.locale = nil` gives a user who wants it.

## Alternative considered: split `LC_MESSAGES` and `LC_CTYPE`

Set `LC_MESSAGES=C` (for the message language) and `LC_CTYPE=C.UTF-8` (for the charset)
instead of a single `LC_ALL`. This looks like it should degrade better on macOS ≤ 14 —
`LC_MESSAGES=C` names a locale POSIX mandates exists, so the message half becomes
guaranteed rather than merely observed. It is also genuinely narrower: `LC_ALL` overwrites
collate, time, numeric, and monetary as well, for git and for every hook, textconv filter,
pager, and credential helper git spawns.

**Rejected, but on narrower grounds than the previous revision claimed.** It does buy
something — it stops imposing `en_US`/`C` time and number formatting on user hooks — but
nothing the gem itself can observe: nothing under `lib/` parses locale-formatted dates or
numbers. What it does *not* buy is any improvement on the availability problem, which is
the whole point: the ctype half carries exactly the same risk and fails exactly the same
way. The second row of the fallback measurement above shows `LC_CTYPE=<nonexistent>`
losing UTF-8 ctype even with `LANG` set to a real UTF-8 locale. And the message half is
already safe under `LC_ALL=C.UTF-8` in practice: git falls back to English when it cannot
load a catalog, which is why message language was English on every platform tested
regardless of value. Two variables instead of one, for a guarantee that only formalizes
what already happens.

## Scope

- **`main` and `4.x`.** The 4.x branch carries the identical pin at `lib/git/lib.rb:2091`
  (verified) and is broken in exactly the same way on Debian, Ubuntu, and RHEL images.
  **Decision: backport**, as Part 2 above.

  A previous revision declined this, on the grounds that the gain did not justify churn on
  a maintenance branch and that a one-line value change with no opt-out is a silent
  behavior change for 4.x users. Both halves of that have been reconsidered:

  - The "no opt-out" objection was specific to the flat `C.UTF-8` proposal, which took
    something away from macOS ≤ 14. The platform-conditional value takes nothing away from
    anyone outside RHEL 7, so there is nothing to opt out of.
  - The churn argument does not survive the bug framing. v5.0.0 shipped 2026-07-28, so
    effectively the entire installed base is on 4.x — the branch where these silently wrong
    answers are actually being returned. Declining to fix it there while calling it a
    data-correctness bug on `main` is not a coherent pair of positions, and the branch
    policy explicitly admits backward-compatible bug fixes to `4.x`.
- **This is a user-visible behavior change** on glibc hosts where `en_US.UTF-8` is not
  generated: it restores `grep`, `log`, and `config` value-regex matches that are silently
  missing today. It changes nothing on macOS, Windows, Alpine, or any glibc host that has
  `en_US.UTF-8` generated. Patch-level on both branches, and it warrants a CHANGELOG entry
  — a silent result change on a patch bump is exactly what a changelog is for, even when
  the change is strictly a fix.

## Follow-up unrelated to the value

The comment above the Non-English Locale job
(`.github/workflows/continuous_integration.yml`) states that runners "generate no locale
beyond C/C.UTF-8". `en_US.utf8` is in fact generated on `ubuntu-latest`. The job's
operational conclusion is unaffected — `de_DE.UTF-8` genuinely is absent and still needs
`locale-gen` — but the sentence is inaccurate and worth correcting when that file is next
touched.

## Context

Came out of reviewing #1667 and fixing #1668 (a spec that inherited the developer's
locale). Neither depends on this issue.

## Revision notes

Revised after an adversarial review of the first revision. What changed:

1. **Corrected the affected-surface table.** The defect is not confined to
   `ignore_case` / `regexp_ignore_case`; case-sensitive patterns containing `.`, a
   quantifier, a POSIX class, or a non-ASCII bracket expression are equally broken, as is
   `perl_regexp`. Replicated on three glibc/git pairs with positive controls. The previous
   control could not have detected this, and the "case-sensitive is fine" conclusion was an
   over-generalization from it.
2. **Reframed around one mechanism** — whether the pinned value yields a UTF-8 ctype —
   rather than around case folding, which is one symptom of several.
3. **Added the escape-hatch deliverable** after verifying that no public API can override
   the pin today.
4. **Corrected the regressing population** from "macOS 14" to "macOS ≤ 14", noted the
   hardware-capped Intel Macs, and flagged the conflict with the stated support policy.
5. **Replaced the proposed container CI job** with a measured `LOCPATH` technique that
   discriminates inside an existing ubuntu job, and replaced "skip loudly" with a
   platform gate.
6. **Led with the determinism argument** (UTF-8 ctype on 15 of 17 platforms after the
   change versus 9 of 17 today) instead of an unmeasurable claim about install counts.
7. **Noted that the pin's original rationale is stale** — `branching.rb` now reads
   `--format`, so the ctype half is the half that matters.
8. **Narrowed the rejection of the `LC_MESSAGES`/`LC_CTYPE` split**: it does buy something
   (it stops imposing time and number formatting on user hooks), just nothing that bears on
   availability.
9. **Split the deliverables** so the fix is not blocked on the regression guard.

### Third revision

Revised again after a second adversarial review, which asked what the surface table was
missing rather than whether its existing rows were right.

1. **Widened the affected-surface table again.** The table was built by reasoning forward
   from `grep`, so it missed every *other* command class that hands a user-supplied
   pattern to git's regex engine: the six `Configuring` methods taking a `value_regex`
   (public API today), and `-G` / `-S --pickaxe-regex` / `--word-diff-regex` on the four
   diff command classes (not yet reachable from the facade, but pending — the diff option
   whitelists are narrow to keep output parsable and to carry inherited limits forward, and
   the intent is to drop the latter). All measured broken under C ctype, all with passing
   controls. This is the second consecutive revision to widen this table, which is itself
   evidence that the enumeration method — not the measurement — was the weak part.
2. **Recorded a second symptom class.** `--word-diff-regex` does not silently drop matches;
   it splits multibyte characters mid-sequence and emits invalid UTF-8.
3. **Replicated the new rows on two glibc/git pairs** — Debian 12 (2.36 / 2.39.5) and
   Ubuntu 24.04 (2.39 / 2.43.0) — each with a second pass after `locale-gen en_US.UTF-8`
   proving the variable is locale availability rather than the value's semantics. Added an
   end-to-end reproduction of the config surface through the public API.
4. **Withdrew a bad probe.** The first `word_diff_regex` measurement counted changed tokens;
   that count is 2 under every locale and discriminates nothing. Replaced with a byte-level
   comparison. Recorded here so it is not reused — it is the same failure shape as the
   metacharacter-free control that caused the previous revision's over-generalization.
5. **Re-scoped the Part 1 coverage and CHANGELOG deliverables** accordingly: testing only
   `grep` and `log` would leave the public config surface unguarded.
6. **Measured the regression instead of inferring it.** The macOS ≤ 14 cost was derived
   from the availability table on the assumption that BSD libc degrades like glibc. Checked
   with a nonexistent locale on macOS 26: it does.

### Fourth revision — decisions recorded

The first three revisions sharpened the evidence but left the design open. This revision
records three decisions, after a review that asked whether the trade-off had to be accepted
at all. The evidence sections are unchanged; what changed is what the issue proposes.

1. **The pinned value is now platform-conditional** — `en_US.UTF-8` on Darwin, `C.UTF-8`
   elsewhere — rather than a flat `C.UTF-8`. This was not considered in earlier revisions.
   It removes the entire macOS trade-off rather than compensating for it: no regressing
   in-policy platform, no CHANGELOG disclosure of losing platforms, no Darwin skip-gate in
   the new spec, and no support-policy question about a macOS floor. The flat value is
   retained as a documented rejected alternative.
2. **The escape hatch is deferred, not shipped.** Its main justification was the macOS
   regression, which no longer occurs. Recorded with the reasons — including the fact that
   the three-state design it needs does not fit the `binary_path` idiom it claimed to copy —
   so that reviving it starts from the design problem rather than rediscovering it.
3. **The fix is backported to `4.x`.** Reversed from "main only". With v5.0.0 eleven days
   old, 4.x is where essentially every affected user is, and the "no opt-out" objection was
   an artifact of the flat-value proposal.

Also dropped the "UTF-8 ctype on 15 of 17 platforms" framing. The number changes under this
proposal, but it should not be used either way: the seventeen platforms are a convenience
sample, not a population, and two rows share a Debian base. And fixed a rendering bug — an
issue reference had wrapped to column 1, where GitHub renders `#295` as a heading.

### Fifth revision — corrected by implementing it

Revised while shipping Part 1 in #1681, after the new coverage turned `windows-latest` red.
The proposal itself is unchanged; what changed is one evidence row and one instruction that
the evidence row had produced.

1. **"Windows is unaffected" is now scoped to the pinned value.** The new metacharacter
   examples fail on Windows — and, measured under `en_US.UTF-8`, `C.UTF-8`, `C`, and no pin
   at all, fail identically under every one of them. So the pin is not the cause and the
   proposal does not regress Windows. But the row as written implied Windows had nothing
   wrong with it, and it does: non-ASCII metacharacter matching does not work there at all,
   independent of this issue. The old row was measured with a folding probe, and the third
   revision widened the surface table without re-running Windows.
2. **"Do not add a platform skip; there is nothing to skip for" is withdrawn.** There is:
   the three metacharacter examples are skipped on `Gem.win_platform?`. The rule it was
   protecting is intact and restated — that gate reads the platform, never the locale, so it
   does not disappear in the environments where the defect lives.
3. **The surface table is now scoped** to glibc, musl, and Darwin, with the Windows regime
   called out separately.

Worth recording as a pattern rather than three separate mistakes: this is the third
consecutive revision in which a probe of one shape was generalized to a surface it cannot
measure — first a metacharacter-free control taken as evidence about metacharacters, then a
changed-token count taken as evidence about tokenizer output, now a folding probe taken as
evidence about `.`. Each was caught by a *different* method than the one that produced it.
The generalizable lesson is the one already in "Method": a probe only licenses claims about
the exact construct it exercised, on the exact platform it ran on.

### Sixth revision — corrected by backporting it

Revised while shipping Part 2 in #1682. The proposal is unchanged in substance; what changed
is how the Darwin branch is *detected*.

1. **`RUBY_PLATFORM.include?('darwin')` is not a sufficient Darwin test.** JRuby reports
   `RUBY_PLATFORM` as `"java"` on every host and records the real platform only in
   `RUBY_DESCRIPTION` (as, for example, `"... [arm64-darwin]"`). So the form this issue
   specified — and that #1681 shipped — put **JRuby on macOS onto the non-Darwin branch**,
   the single platform pairing the Darwin branch exists to prevent. On macOS 11–14 that pins
   a locale the host does not have, degrades to a C ctype, and reproduces every failure in
   the surface table on the platform the branch was added to protect.

   The detection is now
   `RUBY_PLATFORM.include?('darwin') || RUBY_DESCRIPTION.include?('darwin')`, extracted into
   a `darwin_platform?` predicate so the reasoning has somewhere to live and does not get
   "simplified" back. Fixed on `main` in #1684; `4.x` shipped correct in #1682.

2. **This is a supported combination that CI structurally cannot see.** `jruby-10.0.0.1` is
   in both branches' matrices and the gemspec sets no macOS floor, but there is no macOS job
   on either branch and JRuby runs only on `ubuntu-latest`. Recorded under "CI cannot see
   macOS at all", which now has a concrete instance rather than only an argument.

3. **JRuby on Windows was checked and is fine.** `Gem.win_platform?` — which the new spec's
   platform gate uses — reads `RbConfig::CONFIG['host_os']`, and JRuby sets that to the real
   OS (`mswin32` on Windows, `darwin` on macOS) rather than to `java`, so the gate is
   JRuby-safe. JRuby on Windows also lands on the non-Darwin branch, which is correct, and
   the pinned value is measured to make no difference on Windows regardless.

   That fact suggests `RbConfig::CONFIG['host_os'].include?('darwin')` as a single-variable
   alternative to the `||`, and arguably a better one: it is a structured value JRuby
   normalizes for exactly this purpose, rather than a human-readable string being sniffed for
   a bracketed platform tag. **Not adopted** — both branches carry the `||` form — but
   recorded as the first thing to consider if this is ever revisited.

The pattern is worth naming, because it is the same one the fifth revision recorded one level
down. There, probes of one shape were generalized to surfaces they could not measure. Here it
is the *implementation*: `RUBY_PLATFORM` answers "what platform was this Ruby built for",
which is the same question as "what OS is this" on CRuby and a different question on JRuby.
The rule in "Method" — that a measurement licenses claims only about what it actually
exercised — applies to the code just as much as to the evidence.

### LINKED ISSUE #1670: ci: run the test suite under a non-English locale
## Summary

Add a CI job that runs the test suite under a non-English locale, so that specs which
depend on English git output fail in CI rather than on a contributor's machine.

## Motivation

CI never exercises a non-English locale. No workflow sets `LANG` or `LC_ALL`, and
GitHub-hosted runners default to English, so git's translated diagnostics are invisible
here — but they are the *default* experience for a contributor whose shell locale is not
English.

That is not hypothetical. The author of #1667 could not get `bundle exec rake` to pass on
an unmodified `main` and had to export `LC_ALL=en_US.UTF-8` for the whole suite before
they could work. The specific failure was fixed in #1668, but nothing stops the next one
from being introduced, and the failure mode is bad: a first-time contributor sees a red
suite on code they did not touch and has to work out why on their own.

## Proposal

Add one CI job (or matrix entry) that runs the suite with a non-English locale, e.g.:

```yaml
env:
  LANG: de_DE.UTF-8
  LC_ALL: de_DE.UTF-8
```

German is a good choice because git's German translation is complete enough to catch
message-matching assumptions, and it is the locale from the original report in #753.

Note that ubuntu runners do not ship generated locales — the job needs a
`sudo locale-gen de_DE.UTF-8 && sudo update-locale` step (or the `locales` package) first,
otherwise the setting is silently inert and the job proves nothing. Worth asserting in the
job that git actually emits German before running the suite, so a misconfigured job cannot
pass vacuously.

## Blocked by

PR #1668 must land first. Until it does, this job fails on
`spec/integration/git/command_line_spec.rb` — the known instance this issue is designed to
catch.

## Scope and risk

Turning this on may surface further latent locale assumptions beyond the known one. That
open-endedness is why it is filed separately rather than folded into the smaller fixes.

A quick audit suggests the blast radius is small: `spec/integration/git/command_line_spec.rb`
is the only spec that constructs the `CommandLine` layer against a real git binary, and the
other English-string assertions in the suite (`not a git repository`, etc.) live in unit
specs with mocked stderr, so they are locale-immune by construction. Everything routed
through `Git::ExecutionContext` is protected by its `LC_ALL` pin. But the audit was
grep-based, so treat that as an estimate rather than a guarantee.

Relatedly, this job is what would give the pin in `Git::ExecutionContext` real coverage —
see #1669 for the rest of that story.

## Acceptance criteria

- [ ] A CI job runs the full suite under a non-English locale.
- [ ] The job verifies the locale actually took effect before running the suite.
- [ ] Any failures it surfaces are either fixed or triaged into follow-up issues.

### LINKED ISSUE #1674: ci: populate bundler caches at default-branch scope so PRs get cache hits
## Problem

`continuous_integration.yml` triggers only on `pull_request`. GitHub Actions scopes a cache written during a PR run to that PR's ref (`refs/pull/<n>/merge`), and only caches at **default-branch** scope are readable by every branch. Because CI never runs on a push to `main`, it never writes a cache any other PR can read, so nearly every PR pays a cold `bundle install` on nearly every job.

The one exception is Ruby 4.0, and only by accident: [`release.yml`](../blob/main/.github/workflows/release.yml) uses `ruby-version: ruby`, which currently resolves to 4.0.6, and it runs on pushes to `main`. The v5.0.4 release populated that cache.

## Evidence

From the first (cold) run on the `ci/split-lint-job-from-matrix` branch, run `31216830767` attempt 1 — `Setup Ruby` accounted for 293s of 925s total:

| Job | `Setup Ruby` | Cache |
| --- | --- | --- |
| Ruby 3.2 on windows-latest | 109s | miss |
| Ruby jruby-10.0.0.1 on ubuntu-latest | 54s | miss |
| Ruby truffleruby-24.2.1 on ubuntu-latest | 45s | miss |
| Ruby 3.2 on ubuntu-latest | 41s | miss |
| Lint and Docs (Ruby 3.4) | 35s | miss |
| Ruby 4.0 on ubuntu-latest | **9s** | **hit** |

Attempt 2 of the same run, with the PR-scoped caches now warm, took 731s total versus 925s.

## Expected value

If every job matched the cache-hitting 4.0 job's 9s setup, that is roughly **240s off total job-seconds and 100s off the critical path** -- comparable to the savings in PR #1672.

## Options

1. **A dedicated cache-warming workflow on `push: [main]`** that runs `ruby/setup-ruby` with `bundler-cache: true` for each version in the matrix and nothing else. Cheapest: no specs, no lint, just populates the caches at default-branch scope.
2. **Add `push: branches: [main]` to the existing CI workflow.** Simpler diff, but runs the entire matrix on every merge, which partly spends the savings it creates.

Option 1 is likely preferable.

## Implementation note

PR #1672 added a workflow-level concurrency group keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true`. A `push`-triggered run needs either its own concurrency group or `cancel-in-progress: false`, otherwise consecutive merges to `main` will cancel each other's cache warming.

Also worth confirming: cache keys include the runner OS, Ruby version, and `Gemfile.lock` hash, so the warming job must cover every `(os, ruby)` pair the matrix uses, including `windows-latest` + 3.2.

Found while verifying PR #1672.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #1675: ci: decide whether CI should be a required status check for merging
> **Status (2026-08-09):** Decided and implemented for `main` — CI now gates merges there.
> See the [decision](https://github.com/ruby-git/ruby-git/issues/1675#issuecomment-5223023041)
> and [verification](https://github.com/ruby-git/ruby-git/issues/1675#issuecomment-5227118691)
> comments for how it got here.
>
> **One item remains: the `4.x` branch has no working gate.** See
> [Remaining work: 4.x](#remaining-work-4x) below. The original proposal in this body is
> superseded and kept only as history.

## Problem (as originally filed)

CI is not currently a merge gate. The `Release Branch` ruleset (id `11716425`), which covers `~DEFAULT_BRANCH` and `refs/heads/4.x`, lists exactly one required status check:

```
Verify Conventional Commits
```

Its other rules are `deletion`, `required_linear_history`, `pull_request`, and `non_fast_forward`. None of the build jobs are required, so **a pull request with a completely red matrix is mergeable today.**

This may well be deliberate. Opening it as a decision to make explicitly rather than a bug.

## Current state: what the ruleset actually requires

The decision was to gate. The `Release Branch` ruleset now requires five contexts:

| Required context | Job | Workflow |
| --- | --- | --- |
| `Verify Conventional Commits` | `commit-lint` | `enforce_conventional_commits.yml` |
| `Lint and Docs` | `lint` | `continuous_integration.yml` |
| `All Specs Passed` | `build-complete` | `continuous_integration.yml` |
| `Non-English Locale` | `locale` | `continuous_integration.yml` |
| `Missing en_US.UTF-8 Locale` | `missing-locale` | `continuous_integration.yml` |

Two things changed from the original proposal below:

- **The five per-matrix `Ruby X on Y` contexts are gone and cannot come back.** A required check is matched by name, and the matrix job's name is a template that GitHub only expands once it evaluates the matrix — which the job-level `if:` guard short-circuits on release PRs. Requiring them blocked the v5.0.5 release PR indefinitely. They were replaced by the matrix-free `All Specs Passed` aggregator in #1680. Full analysis in the [verification comment](https://github.com/ruby-git/ruby-git/issues/1675#issuecomment-5227118691).
- **Two locale jobs were added afterward** and are also required: `Non-English Locale` (issue #1670) and `Missing en_US.UTF-8 Locale` (issue #1669). Neither existed when this issue was filed.

## Remaining work: 4.x

The ruleset's `ref_name.include` is `["~DEFAULT_BRANCH", "refs/heads/4.x"]` — **one shared list of required checks for two branches whose workflows are not the same.**

`4.x`'s `continuous_integration.yml` has only the `build` matrix job. It has no `lint`, no `build-complete`, and neither locale job. So on a `4.x` pull request, four of the five required contexts never report at all — not skipped, not failed, simply absent — which blocks the PR indefinitely.

This is live today on #1683 (`chore: release v4.4.1`):

```
mergeStateStatus: BLOCKED   rollup state: SUCCESS
reported: "Ruby ${{ matrix.ruby }} on ${{ matrix.operating-system }}"  skipped
          "Verify Conventional Commits"                                skipped
```

Nothing is failing. The four missing contexts sit pending forever. This is the same failure mode that stranded the v5.0.5 release PR, now reproduced on `4.x` — and it means `4.x` currently has no working gate at all, only a deadlock.

Note that `4.x`'s matrix job runs `bundle exec rake default`, which includes RuboCop and YARD. There is no separate lint job to require there, so a single aggregator covers everything that branch's CI does.

### Plan

1. Backport a `build-complete` job named `All Specs Passed` to `4.x`'s `continuous_integration.yml` — `needs: [build]`, `if: always()`, same `success | skipped` check as `main`. No release-please guard on the aggregator itself; that is the point of it. **Open as #1687.**
2. Split the ruleset once, after that job exists and has reported:

   | Branch | Required contexts |
   | --- | --- |
   | `main` | `Verify Conventional Commits`, `Lint and Docs`, `All Specs Passed`, `Non-English Locale`, `Missing en_US.UTF-8 Locale` |
   | `4.x` | `Verify Conventional Commits`, `All Specs Passed` |

The two locale jobs are deliberately not backported: they gate a defect class on `main`, and `4.x` takes only security and backward-compatible bug fixes. Revisit if a locale-sensitive fix is ever backported.

Ordering matters. Splitting first would leave `4.x` unguarded during the gap; backporting first means #1683 stays blocked until the new job reports, which needs a fresh run on that PR (release-please regenerating its branch is usually enough).

## Caveat: release-please PRs

*Resolved on `main` — kept for context. See the [verification comment](https://github.com/ruby-git/ruby-git/issues/1675#issuecomment-5227118691) for what actually happened.*

Both jobs in `continuous_integration.yml` carry a guard that skips them on release branches:

```yaml
if: >-
  github.event_name == 'workflow_dispatch' ||
  (github.event_name == 'pull_request' && !startsWith(github.event.pull_request.head.ref, 'release-please--'))
```

Making these checks required turns that guard load-bearing. A job skipped by a job-level `if` reports a `skipped` conclusion, which rulesets treat as passing, so release PRs should still be mergeable -- but this should be **verified on the first release after enabling it**, since a required check that never reports at all blocks a PR indefinitely.

Note the same consideration applies if a merge queue is ever adopted: `merge_group` events carry no `github.event.pull_request.head.ref`, so the guard above would need rewriting first. This caveat is unaffected by the fix and still applies.

## Original proposal (superseded)

*Kept as history. The per-matrix contexts below are not what shipped — see "Current state" above.*

Add these contexts to the ruleset's `required_status_checks` (names are current as of PR #1672):

- `Lint and Docs`
- `Ruby 3.2 on ubuntu-latest`
- `Ruby 4.0 on ubuntu-latest`
- `Ruby truffleruby-24.2.1 on ubuntu-latest`
- `Ruby jruby-10.0.0.1 on ubuntu-latest`
- `Ruby 3.2 on windows-latest`

### Consideration

Ruby 4.0, truffleruby, and jruby are the slowest jobs. Requiring all six means every merge waits for the full matrix. An alternative is to require only the fast ubuntu jobs plus `Lint and Docs` and leave the alternative runtimes advisory -- though that reintroduces the possibility of merging a JRuby or Windows regression.

Found while verifying PR #1672.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #1676: ci: remove the vestigial experimental and java_version matrix dimensions
## Problem

The `build` job matrix in `continuous_integration.yml` carries three settings that no longer vary, so they add noise without affecting behavior:

```yaml
    continue-on-error: ${{ matrix.experimental == 'Yes' }}
    env: { JAVA_OPTS: -Djdk.io.File.enableADS=true }

    strategy:
      matrix:
        ruby: ["3.2", "4.0", "truffleruby-24.2.1", "jruby-10.0.0.1"]
        operating-system: [ubuntu-latest]
        experimental: [No]
        java_version: [""]
        include:
          - ruby: "3.2"
            operating-system: windows-latest
            experimental: No
            java_version: ""
```

- **`experimental` is always `No`**, so `continue-on-error` always evaluates to `false`. Genuinely experimental runtimes live in `experimental_continuous_integration.yml`, which sets `continue-on-error: true` directly.
- **`java_version` is always `""`**, so the `Setup Java` step's `if: matrix.java_version != ''` never passes and the step never runs.

Both dimensions read as if the matrix has experimental rows or JVM setup when it has neither. A reader has to check every row to discover otherwise.

## Proposed change

Delete `experimental` and `java_version` from the matrix and the `include` entry, delete the `continue-on-error` line, and delete the `Setup Java` step.

`env: { JAVA_OPTS: -Djdk.io.File.enableADS=true }` should be **kept** -- it is unconditional job-level env, not tied to `java_version`, and it is what makes JRuby handle Windows alternate data streams correctly.

Zero behavior change; all six jobs should run exactly as they do now.

## Verification

Confirm the run produces the same job names and the same pass/fail results, and that no job log contains a `Setup Java` step.

Pre-existing; noticed while reviewing PR #1672 but deliberately left out of that diff to keep it focused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jcouball:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* Detect Darwin under JRuby when pinning LC_ALL ([89bf625](https://github.com/ruby-git/ruby-git/commit/89bf62579999bec27a9b0f1aab16d37e7d5af2b8)), closes [#1669](https://github.com/ruby-git/ruby-git/issues/1669)
* Pin LC_ALL to a locale that exists on the host platform ([5a4e0d6](https://github.com/ruby-git/ruby-git/commit/5a4e0d657b0e0828de0b38f283ae60dfbe2d4cc6)), closes [#1669](https://github.com/ruby-git/ruby-git/issues/1669)

### Other Changes

* Correct two imprecise claims in the build-complete comment ([c4391cd](https://github.com/ruby-git/ruby-git/commit/c4391cdf441a8379dee4cda59f800da41bf8e8b2)), closes [#1675](https://github.com/ruby-git/ruby-git/issues/1675)
* Gate merges on a matrix-free job instead of the per-matrix checks ([47f2d17](https://github.com/ruby-git/ruby-git/commit/47f2d17d16d1ea201bf9cc8f74aedffa7619eb0a)), closes [#1675](https://github.com/ruby-git/ruby-git/issues/1675)
* Remove vestigial matrix dimensions ([3a89886](https://github.com/ruby-git/ruby-git/commit/3a8988650a7d8f0a81c093c470fb9f524142c9b9)), closes [#1676](https://github.com/ruby-git/ruby-git/issues/1676)
* Run RuboCop and YARD once instead of on every matrix job ([0726964](https://github.com/ruby-git/ruby-git/commit/072696491662867cfbc0424febcf36542548f051))
* Run the suite once with en_US.UTF-8 unavailable ([6e012c5](https://github.com/ruby-git/ruby-git/commit/6e012c54d52c8b0c25bf629f07e310b6fcbac229)), closes [#1669](https://github.com/ruby-git/ruby-git/issues/1669)
* Run the test suite under a non-English locale ([c65d402](https://github.com/ruby-git/ruby-git/commit/c65d4022f358a4e4e676dbf2492427851205cdd6)), closes [#1670](https://github.com/ruby-git/ruby-git/issues/1670)
* Skip the All Specs Passed gate on release PRs ([6dfeee0](https://github.com/ruby-git/ruby-git/commit/6dfeee0ed508932393106c8887d9b7c9ef9dbbca)), closes [#1675](https://github.com/ruby-git/ruby-git/issues/1675)
* Warm bundler caches at default-branch scope on push to main ([2821477](https://github.com/ruby-git/ruby-git/commit/282147796f6b859fb9c7f8b5a614cab399cc9539)), closes [#1674](https://github.com/ruby-git/ruby-git/issues/1674)

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).


jcouball:
🤖 Created releases:

- [v5.0.5](https://github.com/ruby-git/ruby-git/releases/tag/v5.0.5)

:sunflower:

PR #1705, opened 2026-08-20
ci: add a markdown link checker

jcouball:
Second of four. **Stacked on #1704** — base is `docs-self-contained-command-policy`, so review only the top commit.

## Why

Nothing in the build reads documentation links. RuboCop reads Ruby, YARD reads doc comments, the specs read code. A relative path into a file that has since moved, and an anchor into a heading that has since been reworded, both stay green forever.

That is not hypothetical. An audit this month found a README section announcing a shipped release in the future tense, stale document paths, and a mis-slugged anchor, some over a year old. PR 4 in this stack rewrites roughly ten inbound references plus six internal links, and I would rather that be verified mechanically than by hand.

## What changed

A `Markdown Links` job running [lychee](https://lychee.cli.rs), configured by `.lychee.toml`. It is a separate job rather than another step in `lint` because it needs no Ruby and no bundle — a broken link should report as a broken link, not as a failure of the job that gates RuboCop.

**It runs offline.** Every link this repository controls is either a relative path or an in-document anchor. External URLs fail for reasons that have nothing to do with the commit under test: rate limits, redirects behind a login, sites that reject datacenter IPs. A gate that goes red on someone else's outage gets ignored, and an ignored gate catches nothing.

Fragments are checked, not just paths — half the links here point at a section, so path-only checking would still let a renamed heading rot.

## The twelve broken links this found

Fixed here so the gate starts green (1203 links checked, 0 errors):

- **5** contents entries in `extract-command-from-lib` that slugged an em-dash heading with one hyphen where GitHub emits two — `#step-1-identify-...` should be `#step-1--identify-...`
- **4** paths into skills that moved between `skills/` and `skills-deprecated/`
- **1** anchor into a `facade-implementation` section that no longer exists
- **1** root-relative `/CONTRIBUTING.md`, which resolves against the site root on GitHub and against nothing at all locally
- **1** contents entry in `redesign/info_object_migration_plan.md` pointing at an anchor for a heading since shortened to "Open threads"

## Running it locally

```bash
brew install lychee   # or: cargo install lychee
lychee --config .lychee.toml .
```

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -50,6 +50,49 @@ jobs:
         run: bundle exec rake rubocop yard
         timeout-minutes: 15
 
+  # Ensure that local links in markdown docs are valid.
+  #
+  # The checker is configured in .lychee.toml. Run the same check locally with:
+  #
+  #     rake markdown:links
+  #
+  links:
+    name: Markdown Links
+
+    # Skip this job if triggered by a release PR
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && !startsWith(github.event.pull_request.head.ref, 'release-please--'))
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v6
+
+      # Track the newest lychee release rather than pinning one. Without this the
+      # action installs whatever version its own default names, which moves when the
+      # `@v2` tag moves -- pinned, but pinned by someone else and bumped without a
+      # commit here. `latest` at least makes the choice ours and keeps the checker
+      # current. The action accepts `latest`, `nightly`, or an exact tag; there is no
+      # major-version form, and lychee is 0.x, so a "compatible range" is not on offer.
+      #
+      # This can break the build on a release nobody here asked for. .lychee.toml is
+      # coupled to the lychee version -- `include_fragments` took a bare boolean before
+      # 0.24 and an enum string from 0.24 on -- and a future release is free to change
+      # a key again. That trade is deliberate: an occasional known-cause failure is
+      # cheaper than a checker that silently ages.
+      #
+      # When it does break, read the parse error, fix .lychee.toml against the current
+      # lychee docs, and commit that. Pin `lycheeVersion` to the last good tag only to
+      # unblock an unrelated PR, and open an issue to unpin.
+      - name: Check Markdown Links
+        uses: lycheeverse/lychee-action@v2
+        with:
+          lycheeVersion: latest
+          args: --config .lychee.toml .
+          fail: true
+
   build:
     name: Ruby ${{ matrix.ruby }} on ${{ matrix.operating-system }}
 
--- a/.lychee.toml
+++ b/.lychee.toml
@@ -1,0 +1,29 @@
+# Configuration for lychee, the markdown link checker. See https://lychee.cli.rs.
+
+# Check only what this repository controls. Every link in these files is either a
+# relative path into the repository or an anchor within one of its documents, and
+# both break silently when a file moves or a heading is renamed -- which is exactly
+# what this checker exists to catch.
+#
+# External URLs are deliberately not requested. They fail for reasons that have
+# nothing to do with the commit under test: rate limits, redirects behind a login,
+# and sites that reject datacenter IPs. A gate that goes red on someone else's outage
+# gets ignored, and an ignored gate stops catching the internal breakage it was added
+# for.
+offline = true
+
+# Verify heading anchors, not just file paths. Half the links this repository writes
+# point at a section rather than a whole file, so path-only checking would still let
+# a renamed heading rot unnoticed.
+#
+# This key is version-sensitive: it took a bare boolean before lychee 0.24 and takes
+# one of none/anchor-only/text-only/full from 0.24 on. If this file stops parsing,
+# check the current lychee docs for the accepted form before changing anything else.
+include_fragments = "anchor-only"
+
+no_progress = true
+
+# Include .github, where the agent skills live. Generated and vendored trees --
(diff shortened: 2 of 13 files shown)


.lychee.toml
```diff
@@ -0,0 +1,33 @@
...
+# nothing to do with the commit under test: rate limits, redirects behind a login,
+# and sites that reject datacenter IPs. A gate that goes red on someone else's outage
+# gets ignored, and an ignored gate stops catching the internal breakage it was added
+# for.
+offline = true
+
+# Verify heading anchors, not just file paths. Half the links this repository writes
+# point at a section rather than a whole file, so path-only checking would still let
+# a renamed heading rot unnoticed.
+include_fragments = "anchor-only"
```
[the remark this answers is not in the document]
jcouball: This is inverted — the boolean is what fails, and the string is what lychee 0.24 requires. `include_fragments` takes an enum (`none` | `anchor-only` | `text-only` | `full`), matching the CLI's `--include-fragments[=<none|anchor-only|text-only|full>]`.

Boolean, on the version this job installs:

```
$ lychee --config .lychee.toml .
[ERROR] Error while loading config: Failed to parse configuration file
  TOML parse error at line 26, column 21
     26 | include_fragments = true
        |                     ^^^^
  wanted string or table
```

The value as committed, in real CI rather than locally — run 32427932782, job `Markdown Links`:

```
Downloading from: .../lychee-v0.24.2/lychee-x86_64-unknown-linux-gnu.tar.gz
| Total    | 1530 |
| Errors   | 0    |
```

So the config parses and the job is green.

That said, the comment points at something real: the accepted *type* is version-dependent. It was a bare boolean before 0.24 and became this enum in 0.24, so a config file nobody touched can start failing when the lychee binary changes underneath it.

`lycheeVersion: v0.24.2` is now set explicitly on the action to close that off. The action already defaults to this exact version, but that default lives in the action and `@v2` is a moving tag — a new v2.x release can bump the installed lychee with no commit here. Pinning makes the bump a deliberate edit in this repository. v0.24.2 is the current stable release; the only newer tag upstream is `nightly`.

No change to the `include_fragments` value.

.github/workflows/continuous_integration.yml
```diff
@@ -50,6 +50,59 @@ jobs:
         run: bundle exec rake rubocop yard
         timeout-minutes: 15
 
+  # Documentation links are the one part of this repository that nothing else checks.
```
jcouball: Shorten this comment to:

```
# Ensure that local links in markdown docs are valid.
#
# The checker is configured in .lychee.toml. Run the same check locally with:
#
#     rake markdown:links
#
```

tasks/markdown.rake
```diff
@@ -0,0 +1,48 @@
...
+# a required prerequisite all the same -- bin/setup verifies it, and this task fails
+# rather than skipping when it is missing, because `markdown:links` runs as part of the
+# default task and a check that silently no-ops would make a green `bundle exec rake`
+# mean less than it says. Every platform this project supports has a packaged lychee;
+# see the prerequisites table in CONTRIBUTING.md.
+#
+# No version is asserted here. bin/setup enforces the minimum that .lychee.toml needs,
+# and CI installs the newest release rather than a pinned one, so a local binary is
+# expected to drift from CI's; see the comment on that job for why.
+#
```
jcouball: Remove these commends.

tasks/markdown.rake
```diff
@@ -0,0 +1,48 @@
...
+# A green run here still does not guarantee a green CI job, and the gap is the
```
jcouball: Say "successful" instead of "green"

.lychee.toml
```diff
@@ -0,0 +1,39 @@
+# Configuration for lychee, the markdown link checker run by the "Markdown Links"
```
jcouball: Don't mention the rake task or GitHub workflow. Those details belong in layers above this one.

.github/workflows/continuous_integration.yml
jcouball: Applied verbatim.

tasks/markdown.rake
jcouball: Removed — I read this as the paragraph above the line you marked:

> No version is asserted here. bin/setup enforces the minimum that .lychee.toml needs, and CI installs the newest release rather than a pinned one, so a local binary is expected to drift from CI's; see the comment on that job for why.

It reaches up into `bin/setup` and the workflow, which is the same layering you flagged on `.lychee.toml`. Nothing is lost: `bin/setup` owns the version floor and says so where it enforces it.

If you meant a wider block, say which and I'll cut it.

tasks/markdown.rake
jcouball: Done — both occurrences in this file, not just the one on the line you marked:

> A **successful** run here still does not guarantee a **successful** CI job…

and further up, "would make a **successful** `bundle exec rake` mean less than it says."

I left "green" alone in `CONTRIBUTING.md:245`, `CONTRIBUTING.md:1090`, and `continuous_integration.yml:309` — all pre-existing text this PR does not otherwise touch. Happy to sweep them in a follow-up if you want the wording consistent repo-wide.

.lychee.toml
jcouball: Done. The header is now one line:

```toml
# Configuration for lychee, the markdown link checker. See https://lychee.cli.rs.
```

The workflow job, the local invocation, and the install instructions are all gone — the first two are named by the layers that own them, and CONTRIBUTING's prerequisites table has the install commands per platform.

I applied the same rule further down the file, since it was the identical violation: the `include_fragments` comment said "that is what sets the >= 0.24.0 floor bin/setup enforces -- keep the two in sync" and mentioned how CI installs lychee. It now states only the version-sensitivity of the key itself. The sync pointer still exists in `bin/setup`, which is the correct direction — the layer above reaching down, not this one reaching up.

PR #1805, opened 2026-09-09
fix(errors): raise Git::Error instead of a bare SystemCallError from filesystem calls

### LINKED ISSUE #1804: Two check-then-use `File.read` sites can let a bare `Errno` escape the "raises only `Git::Error`" contract
Hi,

I run independent verification of AI-built codebases (Trenyx; the published engagements are at https://trenyx.io/audits.html) and did a pre-registered static read of ruby-git/ruby-git at 6be93e41, the plan hashed and anchored before reading. The important part first: the injection posture is excellent. Commands go to the process as an argv array, the shared Arguments DSL rejects operand values that start with `-` (e.g. `diff_stats.rb:35`, `diff_path_status.rb:99`, `repository/inspecting.rb:99`, `repository/object_operations.rb:78,123`), and commands place an explicit `--` before operands, the double defense that closes the classic malicious-clone-URL-becomes-`--upload-pack` class. GIT_SSH is built into a per-spawn environment rather than global `ENV`, and status/diff parsing uses `--porcelain=v2 -z` with a proper `core.quotePath` unescaper. Nothing here is a security report.

This is about README:204-207: "The git gem raises only `ArgumentError` or errors that subclass `Git::Error` … Rescue `Git::Error` to catch any runtime error raised by this gem." Two sites can break that with a bare `Errno::ENOENT` / `Errno::EACCES`:

1. `lib/git/path_resolver.rb:180-182`: `resolve_gitdir_pointer` does `return path unless File.file?(path)` and then `File.read(path).strip` with no rescue. If the gitdir-pointer file disappears or becomes unreadable between the check and the read, the `Errno` propagates out of `Git.open`.
2. `lib/git/repository/object_operations.rb:318`: `tag_sha` does `File.read(head).chomp if … && File.file?(head)`, same shape, same gap.

It's a narrow race, so low severity and not a security issue, but it's a real gap in a contract the gem clearly intends to keep: the same file already wraps `Errno::ENOENT` at `path_resolver.rb:112` in another method. A `rescue SystemCallError` around each read that re-raises as a `Git::Error` subclass would close both, and one spec that deletes the file between the check and the read would fence it. Happy to re-check whatever you ship. The pre-registration hash and the full record publish with the write-up, which says you fixed it.

SK  
Trenyx

### LINKED ISSUE #1806: repo_size lets Errno::EACCES escape as a bare SystemCallError
## Summary

`Git::Repository#repo_size` (`lib/git/repository.rb`) walks the repository directory with `Find.find` and calls `File.lstat` on each path. It rescues only `Errno::ENOENT` (for files that disappear mid-traversal). An `Errno::EACCES` from `File.lstat`, or an `Errno::ENOENT` raised by `Find.find` itself when the repository directory is removed between the `directory?` check and the walk, escapes as a bare `Errno` instead of a `Git::Error`.

This is the same class of bug as issue 1804.

## Steps to reproduce

The failing case is a directory that is readable but not searchable. `Find.find` lists its entries, then `File.lstat` on each entry fails:

```ruby
repo = Git.init('/tmp/repo')
File.chmod(0o444, '/tmp/repo/.git/objects')
repo.repo_size
# raises Errno::EACCES rather than Git::Error
```

Mode `0o444` matters. With mode `0` the directory cannot be listed at all, so `Dir.children` fails inside `Find.find`, which swallows the error under its default `ignore_error: true` and skips the subtree. That path does not raise.

## Expected behavior

Raise `Git::Error` (with the original error as `cause`) for errors other than `Errno::ENOENT` on a vanished entry. This matches the contract described in the README ("Errors raised by this gem").

## Suggested fix

Wrap the traversal in `Git::SystemCallGuard.call('Failed to compute the repository size')`, keeping the existing `rescue Errno::ENOENT; next` for entries that vanish during the walk.

## Out of scope

Two related behaviors are not addressed here:

- With mode `0`, `Find.find` silently skips the unreadable subtree and `repo_size` returns a total that is quietly too low. Converting the raised errors to `Git::Error` does not change that.
- `File.lstat` is unguarded in the same way elsewhere. `Dir.pwd` in `Git::PathResolver` and `File.expand_path` in `context_helpers_validate_path` and `root_of_worktree` were also unguarded; `File.expand_path` consults `Dir.pwd` for a relative path, so it raises `Errno::ENOENT` when the process working directory has been removed. Those are fixed in PR #1805 alongside this one.

Refs: #1804, #1805

jcouball:
Closes #1804
Closes #1806

## Problem

The README promises that the gem raises only `ArgumentError` or a `Git::Error` subclass. Several filesystem calls let a bare `SystemCallError` escape. The two `File.read` sites in the report fail on a permission problem alone, no race needed: `File.file?` returns true for a file the process cannot read.

The gap was wider than the two reported sites. `Dir.chdir`, `Dir.mktmpdir`, `Tempfile.create`, the archive staging and rename steps, `FileUtils.rm_r` in `Git.export`, the `File.lstat` traversal in `repo_size` (issue 1806), and the `Dir.pwd`/`File.expand_path` path resolution all had the same problem.

## Change

- Add `Git::SystemCallGuard`, a private helper that rescues `SystemCallError` and re-raises `Git::Error` with the original as `cause`. Helpers that yield to a caller's block use `guard.unguarded` so a `SystemCallError` raised by the caller's own code propagates unchanged.
- Guard every site: the gitdir pointer read in `Git.open`, `chdir`, `with_working`, `with_temp_index`, `with_temp_working`, block-form `cat_file_contents`, `each_conflict` and `conflicts`, `archive`, `Git.export`, `repo_size`, and the path expansion in `Git::PathResolver` and `set_index`/`set_working`.
- `File.expand_path` consults `Dir.pwd` for a relative path, so it fails when the process working directory has been removed. In `root_of_worktree` that `Errno::ENOENT` was previously reported as "git binary not found".
- The `chdir` guard message no longer names the target directory. `Dir.chdir`'s block form fails on either leg, and naming the target was wrong when restoring the previous directory is what failed. The underlying `SystemCallError` names the directory that actually failed.
- `tag_sha` is the one site that does not raise. The loose ref read is a fast path, so on a system error it falls through to `git show-ref`.
- README and the `Git::Error` doc drop the "does not explicitly raise" hedge and say the original `SystemCallError` is available through `cause`. `@raise [Git::Error]` added to each affected public method.

## Behavior change

Code that rescued `Errno::*` directly around the methods above now sees `Git::Error`. This is a bug fix, not a breaking change: the documented contract has always been `Git::Error`, and the bare `Errno` was never part of the public API. Ships in v6.0.0 only; `5.x` is unchanged.

## Tests

Unit specs for each site stub the filesystem call to raise `Errno::EACCES`, `Errno::ENOSPC`, or `Errno::ENOENT` and assert `Git::Error` with the `Errno` as `cause`. Pass-through specs assert a `SystemCallError` raised inside a caller's block is unchanged, and a `chdir` spec asserts the restore-leg failure names the directory that failed rather than the one entered. Two existing archive specs that expected a bare `Errno::ENOSPC` now expect `Git::Error`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,7 @@ the full API of its result.
 
 ## Errors raised by this gem
 
-The git gem raises only `ArgumentError` or errors that subclass `Git::Error`. It
-does not explicitly raise any other types of errors.
+The git gem raises only `ArgumentError` or errors that subclass `Git::Error`.
 
 Rescue `Git::Error` to catch any runtime error raised by this gem, unless you need
 more specific error handling.
@@ -214,6 +213,18 @@ begin
   # some git operation
 rescue Git::Error => e
   puts "An error occurred: #{e.message}"
+end
+```
+
+Operating system errors from the gem's own filesystem operations, such as reading
+a `.git` pointer file or creating a temporary file, are raised as `Git::Error`.
+The original `SystemCallError` is available through `cause`:
+
+```ruby
+begin
+  repo = Git.open('/path/to/repo')
+rescue Git::Error => e
+  puts e.cause.class if e.cause.is_a?(SystemCallError) #=> Errno::EACCES
 end
 ```
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -64,7 +64,9 @@ To prepare:
    your test suite and, if possible, staging.
 6. Fix each deprecation using the entries under
    [Deprecated methods](#deprecated-methods) until the suite is clean.
-7. Upgrade to v6.0.0.
+7. Replace any `rescue Errno::*` around gem calls with `rescue Git::Error` (see
+   [Filesystem errors raised as `Git::Error`](#filesystem-errors-raised-as-giterror)).
+8. Upgrade to v6.0.0.
 
 ### Minimum Ruby version
 
@@ -163,6 +165,40 @@ while still adding the method to `Git::Repository`. That module is gone, so refe
 monkeypatch to an application-owned module and include it into `Git::Repository`. The
 [`Git::Base` removed](#gitbase-removed) entry under "Upgrading to v5.x" shows the
 migration.
+
+### Filesystem errors raised as `Git::Error`
+
+The gem's own filesystem calls used to let a bare `SystemCallError` escape, which
+contradicted the documented contract that the gem raises only `ArgumentError` or a
+`Git::Error` subclass. Those calls now raise `Git::Error`, with the original error
+available through `cause`.
+
+The affected methods are `Git.open` (reading a gitdir pointer file), `Git.export`,
+`Git::Repository#chdir`, `#with_working`, `#with_temp_index`, `#with_temp_working`,
+`#cat_file_contents`, `#each_conflict`, `#conflicts`, `#archive`, and `#repo_size`.
+
+```ruby
+# v5.x
+begin
+  Git.open('/path/to/repo')
+rescue Errno::EACCES => e
+  handle_permission_problem(e)
+end
+
+# v6.x
+begin
+  Git.open('/path/to/repo')
+rescue Git::Error => e
+  raise unless e.cause.is_a?(Errno::EACCES)
+
+  handle_permission_problem(e.cause)
+end
+```
+
+Code that already rescues `Git::Error` needs no change. A `SystemCallError` raised by
+your own block inside `chdir`, `with_working`, `with_temp_index`, `with_temp_working`,
+or the block form of `cat_file_contents` still propagates unchanged, so the gem never
+relabels an error raised by your code.
 
 ---
 
--- a/lib/git.rb
+++ b/lib/git.rb
@@ -89,6 +89,7 @@ require 'git/status_info'
 require 'git/stash'
 require 'git/stash_info'
 require 'git/stashes'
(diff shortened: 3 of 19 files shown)


blu400codes:
Re-checked against 4fbd35a, as offered on #1804. It checks out:

- SystemCallGuard.call re-raises inside the rescue, so cause carries the original Errno; the unguarded region lets a caller's own SystemCallError through unchanged (system_call_guard.rb).
- Every remaining filesystem call in lib/ that can raise is inside a guard: export (git.rb:291), the gitdir pointer read (path_resolver.rb:188), chdir and mktmpdir (context_helpers.rb:267, 283), the merge stage tempfile (merging.rb:450), cat_file_contents (object_operations.rb:88) and archive (object_operations.rb:640, with the tempfile, stat and rename helpers inside the guarded block). The predicates and the force-mode FileUtils calls don't raise, so nothing is left over.
- loose_tag_sha rescues SystemCallError to nil and tag_sha falls through to show-ref.
- Each spec file stubs EACCES at the site it covers, and the README documents cause.

Thanks for widening it to the whole class; that's a better fix than the two sites I reported. The write-up will say v6.0.0 closes it.

SK  
Trenyx

PR #1875, opened 2026-09-18
docs(skills): add the rspec-integration-testing-standards skill

### LINKED ISSUE #1874: Add an rspec-integration-testing-standards skill
## Problem

There is no single set of rules for specs under `spec/integration/`. Guidance is split across three places, and none of them covers the whole suite:

- `.github/skills/command-test-conventions/SKILL.md` (Integration tests section) covers only `spec/integration/git/commands/`.
- `.github/skills/facade-test-conventions/SKILL.md` (Integration tests section) covers only `spec/integration/git/repository/`.
- `CONTRIBUTING.md` (Unit tests vs integration tests) has six high-level bullets and no structural rules.

`.github/skills/rspec-unit-testing-standards/SKILL.md` states that it applies to `spec/unit/` only. Neither convention skill says which of its 28 rules carry over to integration specs, so anyone auditing or writing an integration spec has to invent that mapping. Nothing at all covers `spec/integration/git/parsers/` or the cross-cutting files directly under `spec/integration/git/` (`git_spec.rb`, `command_line_spec.rb`, `load_spec.rb`, `non_ascii_regex_matching_spec.rb`, `thread_safety_spec.rb`, `git_configure_spec.rb`, `url_spec.rb`).

An audit of all 155 integration spec files against the unit standards (read with an ad hoc translation table) found deviations in 117 files. The largest groups are consistent with no guidance reaching them: 63 `raise_error(Git::FailedError)` assertions with no message pattern, 34 command specs using `describe` for `when the command succeeds` / `when the command fails`, 22 `Git.init` calls without `initial_branch:`, about 30 assertions of the `be_a(String)` / `not_to be_empty` / `respond_to` form where the exact value is known, and four specs that cite a "Rule 22 version-variance exception" as a reason to omit the message check entirely, which the rule does not permit.

## Proposal

Add `.github/skills/rspec-integration-testing-standards/SKILL.md` as a standalone skill, not a diff on the unit standards.

Standalone because:

- An agent auditing `spec/integration/` should load one file, not the unit skill plus a translation table. The policy-consolidation experiment (single authority plus links) already showed that linked criteria get skipped.
- Several rules are different rules, not tweaks. Determinism means pinning git config and branch names rather than freezing time. Stubbing is forbidden rather than required. Coverage is not measured. "Assert observable behavior" becomes "assert the value the test set up, not that git printed something." Expressing those as deltas invites misreadings like the "exception" cited above.
- The command and facade convention skills already restate rules in full rather than diffing. A standalone integration skill matches that pattern and gives both of them a base to point at.

Keep the rules that are genuinely identical to the unit skill short and say once at the top that they are meant to match the unit skill word for word, so a reviewer can diff them. Do not reuse the unit rule numbers; independent numbering lets each skill evolve and stops "Rule 18 but for integration" reasoning.

## Content the skill should cover

Scope: every file under `spec/integration/`, including parsers and the cross-cutting files.

Rules carried over unchanged from the unit skill (state them, do not link): constant `RSpec.describe` per class; one `describe` per public method with `context` for conditions prefixed when/with/without; `frozen_string_literal`; `require 'spec_helper'` plus only the file under test; `described_class`; named `subject` first, then `let` defaults, no `subject` overrides in nested contexts; `let` for values and `before` for side effects, no `before(:all)` or instance variables in spec files; `it` asserts one concept and the description matches the assertion; error assertions name the class and a message pattern; no assertions that only prove the code ran (`be_a`, `not_to be_nil`, `not_to be_empty`, `respond_to`, `not_to raise_error`) when the value is known; `change` matchers instead of manual before/after; no `allow_any_instance_of` or `receive_message_chain`.

Rules that are different for integration specs:

- Real git is the point. Do not stub collaborators. A stub in an integration spec needs an inline comment (the `FileUtils.rm_r` stub in `git_spec.rb` is the model).
- Determinism means independence from the host: every `Git.init` passes `initial_branch:`; every repository goes through `init_test_repo` or `pin_test_repo_config` so `core.autocrlf`, `core.editor`, `commit.gpgsign`, and user identity are pinned; no `sleep`; no dependence on the machine's global git config.
- Error assertions anchor on something the test controls (the input value, the ref name, the path) rather than git's phrasing. The version-variance exception loosens the pattern; it never removes it.
- Assert the value the test set up (`eq('main')`, `contain_exactly(...)`, `have_attributes(...)`), not the shape of git's output.
- Coverage is not measured for integration specs. Do not write examples to cover lines.
- Setup git calls go through `execution_context.command_capturing` or `repo.*`, never a sibling command class, `system`, backticks, or `%x`.
- Cross-platform paths only: `File.join`, `Dir.mktmpdir`, `Dir.tmpdir`; never `/tmp`, `/dev/null`, or `/nonexistent/...`.
- Global state (`ENV`, `Dir.chdir`) is block-scoped or restored in `after`.
- Spec location mirrors the source file under `spec/integration/git/`; cross-cutting specs with no source counterpart need a header comment saying what they cover and why they live there.

Directory-specific rules stay in the command and facade convention skills. The new skill points at them for grouping (`when the command succeeds` / `when the command fails` under `#call`), the one-error-test-per-command requirement, `unless_git` guards, and the facade rule that single-command delegators do not get integration tests.

## Follow-up

Once the skill exists, update `rspec-unit-testing-standards`, `command-test-conventions`, `facade-test-conventions`, `pr-readiness-review`, `pull-request-review`, and `CONTRIBUTING.md` to reference it, and file the cleanup of the existing deviations as separate issues grouped by rule.

jcouball:
## Summary

Adds `.github/skills/rspec-integration-testing-standards/SKILL.md`, a standalone
skill covering every spec under `spec/integration/`: command, facade, parser, and
the cross-cutting specs directly under `spec/integration/git/`.

Per issue 1874, the skill uses its own rule numbers rather than the unit skill's.
Rules 1 through 12 restate the unit rules that carry over unchanged and say so at
the top, so a reviewer can diff them. Rules 13 through 21 are the rules that differ
for integration specs:

- real git, no stubbed collaborators; a stub needs an inline comment (the
  `FileUtils.rm_r` stub in `git_spec.rb` is the model)
- host independence: `initial_branch:` on every `Git.init`, every repository
  through `init_test_repo` or `pin_test_repo_config`, no `sleep`, no reliance on
  global git config
- error patterns anchored on test-controlled input; version variance loosens the
  pattern and never removes it
- assert the value the test set up, not the shape of git's output
- no coverage-driven examples
- setup git calls through `execution_context.command_capturing` or `repo`
- cross-platform paths only
- block-scoped or restored global state
- mirrored spec location, with a header comment for cross-cutting specs

Directory-specific rules (command grouping, one error test per command,
`unless_git` guards, facade single-delegator exclusion) stay in the command and
facade convention skills, which the new skill points at.

## Not in this PR

Per the issue's follow-up section: updating `rspec-unit-testing-standards`,
`command-test-conventions`, `facade-test-conventions`, `pr-readiness-review`,
`pull-request-review`, and `CONTRIBUTING.md` to reference the new skill, and
filing the cleanup of existing deviations grouped by rule.

## Verification

- `bundle exec rake markdown:links` passes
- every Contents anchor resolves to a heading in the file
- the Ruby examples use the v6 API (`branch_list`, `Git::BranchInfo`,
  `StatusInfo#added?`)

Closes #1874


jcouball:
Reviewed and ready to merge

PR #1876, opened 2026-09-18
fix(factories)!: expand ~ in the Git.init and Git.clone paths before git runs

### LINKED ISSUE #1873: Git.init and Git.clone hand a ~-prefixed directory to git unexpanded
## Problem

The factory methods disagree on what a `~`-prefixed path means.

`Git.open`, `set_index`, and `set_working` resolve paths in Ruby through `File.expand_path`, so `~/repo` means `$HOME/repo`. `Git.init` and `Git.clone` pass the positional directory to git as given. ruby-git spawns git without a shell and git does no tilde expansion on its command-line arguments, so `~/repo` makes git create a directory literally named `~` under the process working directory (or under `:chdir` for `Git.clone`).

For `Git.init` this is a bug, not just a surprise. `run_init_command` runs `git init` on the literal path, then `open_after_init` calls `Git.open`, which expands `~` to the home directory. The returned repository points at a directory git never created, so the first command run on it fails.

```ruby
repo = Git.init('~/scratch')
# git created ./~/scratch; repo.dir is $HOME/scratch
repo.status  # fails: not a git repository
```

`Git.clone` only works by accident: `resolve_paths_from_clone_result` re-reads the literal path from git's stderr and joins it onto `:chdir`, so the repository is bound to the `~` directory git created.

Related: PR 1872 (issue 1871) already expands the clone `:index` with `File.expand_path(index, chdir)`. The clone directory could not get the same treatment there because git had already created the literal path by the time Ruby resolved it. This issue closes that gap on git's side of the boundary by expanding before git runs.

`Git.init` has the same mismatch on its `:repository` option. `run_init_command` hands `options[:repository]` to `git init --separate-git-dir` as given, and `open_after_init` then opens with the same string through `File.expand_path`. So `Git.init('wt', repository: '~/sep.git')` makes git create `./~/sep.git` while the returned repository points at `$HOME/sep.git`.

## Proposed fix

Expand the positional directory in Ruby before handing it to git, so every factory method agrees with `Git.open` and with Ruby file semantics:

- `Git.init(directory)`: `File.expand_path(directory)` before `git init`.
- `Git.init(directory, repository: path)`: `File.expand_path(path)` before it is passed as `--separate-git-dir`, so git creates the directory the returned repository is bound to.
- `Git.clone(url, directory, chdir: ...)`: `File.expand_path(directory, chdir)` before `git clone`. When no directory is given, git derives the name from the URL and there is nothing to expand.

Route every expansion through `Git::SystemCallGuard` so a removed process working directory raises `Git::Error`, matching the rest of path resolution. Two guarded helpers already exist and should be consolidated into one rather than joined by a third:

- `PathResolver.expand_path(path, message)` is a `private_class_method` with no base directory argument, so it cannot do the `File.expand_path(directory, chdir)` that `Git.clone` needs.
- `Git::Factories#expand_against_chdir(path, chdir)`, added by PR 1872 for the clone `:index`, takes a base but hardcodes the index error message.

Give `PathResolver.expand_path` an optional base directory parameter, make it public with an `@api private` tag, and have `Git.init`, `Git.clone` (both the directory and the `:index`), and the existing `resolve_*` methods all call it. Remove `expand_against_chdir` once nothing uses it.

Passing an absolute path to git changes nothing else: `git clone` still reports the path in its stderr and `parse_clone_stderr` still works. `prefix_with_chdir` still works too, since it passes an absolute path through untouched, but its doc comment says the clone directory is joined rather than expanded because git created it as given. Reword that comment: the directory is already expanded before git runs, and the join exists only for the case where no directory was given and git derived the name from the URL.

Specs that assert on `$HOME` set `ENV['HOME']` to a temporary directory in an `around` hook, restoring the original value afterward, and assert against `Dir.home`. No spec support does this today. Do not rely on the developer's real home directory. The existing clone `:index` spec in `spec/unit/git/factories_spec.rb` that asserts against `Dir.home` without overriding it should use the same hook.

## Decision record

This issue and issue 1871 together settle how the gem treats a path that locates a repository. Record that as an ADR in the same PR, written as the general rule rather than as a summary of the two issues:

- A path a caller gives to a factory method means what it would mean typed at a shell prompt in the process working directory at the time of the call, or in `:chdir` for `Git.clone`. A relative path is expanded against that base and nothing else: never against the working tree or the git directory, which only supply the defaults when an option is omitted (`:repository` defaults to `<working_dir>/.git` and `:index` to `<git_dir>/index`). An absolute path is used as given. The base is `Dir.pwd` at the moment the factory runs, so the same call resolves differently after `Dir.chdir` or inside a `Git::Repository#chdir` block. A relative `:chdir` is itself resolved against the process working directory, since that is where git is started from.
- The two exceptions and why each exists: `Git.clone`'s `:repository` is passed through untouched because git resolves it relative to where git runs; a relative target inside a `gitdir:` pointer file resolves against the pointer file's directory because it is file content and never saw a shell.
- `~` is expanded in Ruby before git sees the path. This follows from the shell-prompt reading of the rule: the shell expands `~` before git runs, and git itself never does. `File.expand_path(path, base)` is the primitive because it does exactly what the shell plus git would do: expands `~`, joins a relative path onto the base, normalizes `.` and `..`, and leaves an absolute path alone. Joining with `File.join` is rejected because it does none of that and glues the base onto an absolute path.
- The rejected alternative: layout-relative resolution (`:repository` against the working directory, `:index` against the git directory), which issue 1871 removed, and why.
- The rule covers only paths that locate a repository. Paths passed to operations such as `add` stay relative to the repository working directory, as the Project Context skill already says. The ADR must say so, or it will appear to contradict that skill. The skill states the operation-path rule in a single line, so the ADR is the fuller statement of the distinction; do not expand the skill to match.

The ADR's number is one past the highest in `docs/adr/` when the PR is opened.

## Scope

The only behavior that changes is a caller who passed a `~`-prefixed directory or a `~`-prefixed `:repository` to `Git.init`. Nobody does that to get a directory named `~`, but it does change what path `Git.clone` creates, so this ships in the next major with a short `UPGRADING.md` entry under v6.x, next to the entry for relative `:repository` and `:index` paths from issue 1871.

Out of scope: `:repository` on `Git.clone`, which git resolves and which stays as documented in issue 1871.

## Acceptance criteria

- [ ] `Git.init('~/scratch')` creates and returns a repository at `$HOME/scratch`
- [ ] `Git.init('wt', repository: '~/sep.git')` creates the git directory at `$HOME/sep.git` and returns a repository whose `repo` is that directory
- [ ] `Git.clone(url, '~/scratch', chdir: dir)` creates and returns a repository at `$HOME/scratch`
- [ ] `Git.clone(url)` with no directory is unchanged
- [ ] A removed process working directory raises `Git::Error` from `Git.init` and `Git.clone`
- [ ] UPGRADING.md documents the change under v6.x
- [ ] An ADR records the path rule, its exceptions, the tilde decision, and the rejected layout-relative alternative, and distinguishes repository-locating paths from operation paths

Refs: #1871

jcouball:
Closes #1873

## Problem

`Git.init` and `Git.clone` handed the positional directory and the `:repository` option to git as given. ruby-git spawns git without a shell and git does no tilde expansion on its arguments, so a `~`-prefixed path made git create a directory literally named `~` under the process working directory, or under `:chdir` for `Git.clone`. `Git.init` then opened the path through `File.expand_path`, so the returned repository pointed at a directory git never created and the first command run on it failed. A `~`-prefixed `:repository` option of `Git.clone` made git fail with an invalid path error.

## Change

- `Git::PathResolver.expand_path` is public (still `@api private`) and takes a `base:` keyword, so one guarded helper serves every expansion. The `expand_against_chdir` helper in `Git::Factories` is removed.
- `Git.init` expands the directory and `:repository` before running `git init`.
- `Git.clone` expands the directory and the `:repository` option against `:chdir`, or the process working directory, before running `git clone`. The `:index` option goes through the same helper. `prefix_with_chdir` is unchanged in behavior and its comment now says the join exists for the URL-derived name case.
- A removed process working directory raises `Git::Error` from `Git.init` and `Git.clone` before git runs.
- ADR-0010 records the rule issues 1871 and 1873 settled, its two exceptions, the tilde decision, the rejected layout-relative alternative, and the distinction from operation paths.
- `UPGRADING.md` documents the change under v6.x next to the relative `:repository` and `:index` entry.

## Tests

- New shared context `with a temporary home directory` points `HOME` at a temporary directory in an `around` hook and restores it. The existing clone `:index` spec that asserted against `Dir.home` now uses it.
- Unit specs cover tilde and relative expansion for the `Git.init` directory and `:repository` and for the `Git.clone` directory and `:repository` with and without `:chdir`, plus the `Git::Error` path for each.
- Integration specs create real repositories at `~/scratch` and `~/sep.git` under the temporary home and assert no `~` directory appears in the process working directory or `:chdir`.

One existing unit example changed: `Git.clone` with a relative directory now asserts that git receives the path expanded against the process working directory, rather than the relative string. That is the behavior the issue asks for.


jcouball:
Reviewed and ready to merge.

PR #1877, opened 2026-09-18
feat(repository): deprecate Git::Repository#set_index and #set_working

### LINKED ISSUE #1833: Deprecate Git::Repository#set_index and #set_working in favor of constructing a bound repository
## Summary

`Git::Repository#set_index` and `#set_working` reassign the receiver's `@execution_context`. Issue 1830 removes the same mutation from the block-taking `with_*` methods because the swap is visible to every other holder of the repository object (a `Git::Object`, a `Git::Branch`, a second caller, another thread). That reasoning applies to `set_index` and `set_working` equally. They stay in v6.0.0 for v4.x compatibility, and this issue deprecates them so a later major can remove them.

## Replacement

A repository bound to a different index or working tree is built up front instead of rebound after the fact:

```ruby
# instead of
repo = Git.open('/path/to/repo')
repo.set_index('/path/to/custom.index')

# build the bound repository
repo = Git.open('/path/to/repo', index: '/path/to/custom.index')

# instead of
repo.set_working('/path/to/worktree')

# build the bound repository
repo = Git.open('/path/to/worktree', repository: '/path/to/repo/.git')
```

`Git.open` already accepts `:index` and `:repository`, so no new API is needed. A caller that needs a bound copy of an existing object for the duration of a block uses `with_index` or `with_working` and the block parameter.

## What to build

- Both methods emit a `Git::Deprecation` warning that names the `Git.open` replacement. Per [ADR-0007](https://github.com/ruby-git/ruby-git/blob/main/docs/adr/0007-removals-require-one-normal-release-of-deprecation-not-calendar-soak.md) the warning says "will be removed in a future major release" until a roadmap decision names the major.
- The YARD docs on both methods carry a `@deprecated` tag pointing at `Git.open`.
- The private `context_helpers_rebuild_context` helper is the only remaining mutation of `@execution_context`. It goes with the methods when they are removed.
- An UPGRADING.md entry in the section for the release that ships the warning shows the replacement.

## Blocked by

- #1830, so the `with_*` methods stop calling `set_index` and `set_working` before those methods warn. Otherwise every `with_*` call would emit the deprecation.
- #1776 removes the positional `check` argument from both methods in v6.0.0. Ship the warning after that removal so the message and signature do not change twice.

## Acceptance criteria

- [ ] `set_index` and `set_working` warn once per call through `Git::Deprecation` with the replacement named
- [ ] Unit specs cover the warning and that the methods still rebind the receiver
- [ ] No internal caller of either method remains in `lib/`
- [ ] YARD `@deprecated` tags on both methods
- [ ] UPGRADING.md entry showing the `Git.open` replacement

jcouball:
Closes #1833

## Summary

`Git::Repository#set_index` and `#set_working` now warn on every call through `Git::Deprecation`. Both methods reassign the receiver's execution context, and the swap is visible to every other holder of the repository object. Issue 1830 removed the same mutation from the `with_*` helpers for that reason. The methods stay in v6.0.0 for v4.x compatibility so a later major can remove them.

Each warning names the replacement: open a repository bound to the index or working tree up front with `Git.open(dir, index:)`, `Git.bare(git_dir, index:)`, or `Git.open(work_dir, repository:)`. The message says "a future major release" rather than a version because `Git::Deprecation`'s horizon is 6.0.0, the release the warning ships in, so `deprecate_methods` would generate the wrong text.

## Changes

- `lib/git/repository/context_helpers.rb`: both methods emit the warning before validating the path, so a caller sees it even when the call raises `ArgumentError`. Both carry a `@deprecated` YARD tag.
- `spec/unit/git/repository/context_helpers_spec.rb`: each describe block gains an example asserting the warning and one asserting it is emitted before path validation. The existing examples move into a nested context silenced by one `around` hook. The `with_index` example that called `set_index` on the receiver is removed; the neighboring examples cover that contract without the deprecated method.
- `UPGRADING.md`: new entry in the v6.x section showing the string-literal, rebind-from-an-existing-object, and block forms of the replacement, which settings each form keeps, the `Git.bare(git_dir, index:)` form for bare repositories, and the `must_exist: false` gap accepted under ADR-0007. The `with_*` paragraph now notes the methods are deprecated.

## Verification

- `bundle exec rspec`: 6524 examples, 0 failures, 4 pending
- `bundle exec rubocop`: no offenses
- `bundle exec yard-lint lib/git/repository/context_helpers.rb`: only the pre-existing `Git::Repository` reopen warning


jcouball:
Reviewed and ready to merge.

PR #1878, opened 2026-09-18
feat(factories)!: raise ArgumentError when Git.open or Git.bare receives an unknown option key

### LINKED ISSUE #1869: Git.open and Git.bare raise ArgumentError on unknown option keys
## Summary

`Git.open` and `Git.bare` ignore option keys they do not recognize. A typo (`Git.open(dir, respository: git_dir)`) or an unsupported key (`Git.bare(dir, index: path)`) returns a repository bound to paths the caller did not ask for, with no signal. For `:repository` that means auto-detecting a working tree the caller meant to override; for `:index` it means every write goes to the repository's live index instead of the scratch one.

`Git.ls_remote` already raises `ArgumentError, "Unknown options: ..."` for unrecognized keys, `Git.clone` forwards unrecognized keys to `Git::Commands::Clone`, which raises, and every `Git::Commands::*` class rejects unknown options. `Git.open` and `Git.bare` are the outliers. ADR-0003 delegates validation of git semantics to git; option keys are the gem's own API contract, so it does not apply.

v4.x and v5.x also ignored unknown keys, so this is a behavior change with no deprecation path. Per the branch policy it ships on `main` in v6.0.0 with an UPGRADING.md entry and is not backported: a v5.x deprecation warning would gate nothing under ADR-0007, which requires a prior deprecation for removals, not for tightened validation.

## Current behavior

```ruby
repo = Git.bare('/srv/repo.git', index: '/tmp/scratch.index')
repo.index #=> #<Pathname:/srv/repo.git/index>

repo = Git.open('/path/to/worktree', respository: '/srv/repo.git')
repo.repo #=> whatever `git rev-parse` finds from /path/to/worktree
```

`from_paths` merges the resolved paths over the caller's options and `Git::ExecutionContext::Repository.from_hash` reads only the keys it knows, so extra keys are dropped rather than rejected.

## What to build

- `Git.open` accepts `:repository`, `:index`, `:log`, `:git_ssh`, and `:binary_path`. `Git.bare` accepts `:log`, `:git_ssh`, `:binary_path`, and `:index` (added by #1868). Any other key raises `ArgumentError, "Unknown options: <keys>"`, the same wording `Git.ls_remote` uses.
- Define the allowed keys as `private_constant`s in `Git::Factories` (`lib/git/factories.rb`), one per factory, following the pattern of `LS_REMOTE_ALLOWED_OPTS` in `lib/git.rb`. Check them before any filesystem or git call so a bad key fails fast.
- `Git.init` already passes filtered hashes to `open` and `bare` through `base_open_options_after_init` and `worktree_open_options_after_init`, so no internal caller changes. Confirm with a spec that `Git.init` with every supported option still opens cleanly through both paths.
- YARD on both factories: add `@raise [ArgumentError] if \`options\` contains an unknown key`, matching `Git.ls_remote`.
- Unit specs in `spec/unit/git/factories_spec.rb` for the raise on each factory, and that each allowed key is still accepted.
- Commit as `feat(factories)!:` so release-please records it as a breaking change.
- UPGRADING.md entry in the v6.x section, near "`Git::Repository` option shims removed", stating that unknown keys now raise, listing the accepted keys per factory, and noting the fix is to drop the key.

## Sequencing

Land after #1868 so the `Git.bare` list is written once with `:index` in it. That is an ordering preference; this issue does not block #1833 or #1868, and neither blocks this one in a way that changes the design.

## Acceptance criteria

- [ ] `Git.open(dir, bogus: 1)` and `Git.bare(dir, bogus: 1)` raise `ArgumentError` naming the key
- [ ] Every documented key on each factory still works
- [ ] `Git.init` with any supported option still opens through both the worktree and bare paths
- [ ] YARD `@raise` tags on both factories
- [ ] UPGRADING.md entry

jcouball:
Closes #1869

## Summary

`Git.open` and `Git.bare` ignored option keys they did not recognize. A typo such as
`respository:` or an unsupported key such as `index:` on `Git.bare` silently returned a
repository bound to paths the caller did not ask for.

Both factories now check the option keys before any filesystem or git call and raise
`ArgumentError, "Unknown options: <keys>"`, the wording `Git.ls_remote` already uses.

Accepted keys:

- `Git.open`: `:repository`, `:index`, `:log`, `:git_ssh`, `:binary_path`
- `Git.bare`: `:index`, `:log`, `:git_ssh`, `:binary_path`

## Changes

- `lib/git/factories.rb`: `OPEN_ALLOWED_OPTS` and `BARE_ALLOWED_OPTS` private constants, a
  private `reject_unknown_option_keys!` helper, the check as the first statement of `open` and
  `bare`, and `@raise [ArgumentError]` YARD tags on both.
- `spec/unit/git/factories_spec.rb`: raise specs for each factory (the message names the keys
  and the check runs before the directory check or path resolution), and `Git.init` with every
  supported option through both the worktree and bare paths.
- `UPGRADING.md`: v6.x entry after "`Git::Repository` option shims removed".

## Notes

- v4.x and v5.x ignored unknown keys, so this is a behavior change with no deprecation path.
  It ships in v6.0.0 and is not backported.
- The helper takes `options.keys` rather than the hash because yard-lint requires `@option`
  tags on any parameter named `options`.


jcouball:
Reviewed and ready to merge.

